PR #5, opened 2014-10-09
WIP: Lazy exists?

Iristyle:
This is a bit of a rework on `instances` which I think is problematic in it's current incarnation.

This works better for me locally, and avoids the problem of querying across multiple regions to find resources up-front.   However, it breaks `puppet resource` and could have other problems.

The `exists?` check does provide a good spot for us to blow up if we find multiple matches in a region.

Food for thought...


Iristyle:
Approach abandoned

PR #19, opened 2014-11-04
Initial support for VPC resources

garethr:
* VPC
* Subnet
* Internet gateway
* Route table

Also includes modifications to instances and security groups to work
with VPC.

I've been using this scenario as a basis to get to the first working
version of this support:
http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/VPC_Scenario1.html

At this point it would be useful to review I think from a functionality
point of view. It's a bigger than ideal diff and it would be better to
get it in sooner rather than later.


garethr:
Closing in favour of work on https://github.com/puppetlabs/puppetlabs-aws/pull/43

PR #30, opened 2014-11-11
autoscaling group support

garethr:
* This adds a type and provider for autoscaling groups and launchconfigs
* This currently supports only the required properties
* It also only supports non-VPC environments at the moment

### DIFF
--- a/examples/auto-scaling-groups/README.md
+++ b/examples/auto-scaling-groups/README.md
@@ -1,7 +1,7 @@
-# Autscaling Support
+# Autoscaling Support
 
 This example demonstrates managing a higher level infrastructure
-component - auto scaling groups. It roughky follows [this
+component - auto scaling groups. It roughly follows [this
 example](http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/policy-creating-cli.html)
 from Amazon.
 
--- a/examples/auto-scaling-groups/README.md
+++ b/examples/auto-scaling-groups/README.md
@@ -27,12 +27,12 @@ directory run:
 
 
 This should create the AWS resources. You can delete them
-with the the destory manifest when done.
+with the the destroy manifest when done.
 
     puppet apply destroy.pp --test
 
 Note that this will shut down any instances launched by the auto scaling
-grup. Once the instances are terminated you can remove the accompanying
+group. Once the instances are terminated you can remove the accompanying
 security group with:
 
    puppet resource ec2_securitygroup test-sg region=sa-east-1 ensure=absent
--- a/examples/auto-scaling-groups/README.md
+++ b/examples/auto-scaling-groups/README.md
@@ -27,15 +27,15 @@ directory run:
 
 
 This should create the AWS resources. You can delete them
-with the `puppet resource` commands when done.
+with the the destory manifest when done.
+
+    puppet apply destroy.pp --test
+
+Note that this will shut down any instances launched by the auto scaling
+grup. Once the instances are terminated you can remove the accompanying
+security group with:
 
    puppet resource ec2_securitygroup test-sg region=sa-east-1 ensure=absent
-   puppet resource ec2_launchconfiguration test-lc region=sa-east-1 ensure=absent
-   puppet resource ec2_autoscalinggroup test-asg region=sa-east-1 ensure=absent
-   puppet resource ec2_scalingpolicy scaleout region=sa-east-1 ensure=absent
-   puppet resource ec2_scalingpolicy scalein region=sa-east-1 ensure=absent
-   puppet resource cloudwatch_alarm AddCapacity region=sa-east-1 ensure=absent
-   puppet resource cloudwatch_alarm RemoveCapacity region=sa-east-1 ensure=absent
 
 
 ## Discussion
--- a/examples/auto-scaling-groups/README.md
+++ b/examples/auto-scaling-groups/README.md
@@ -1,7 +1,9 @@
 # Autscaling Support
 
 This example demonstrates managing a higher level infrastructure
-component - auto scaling groups.
+component - auto scaling groups. It roughky follows [this
+example](http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/policy-creating-cli.html)
+from Amazon.
 
 ## What
 
--- a/examples/auto-scaling-groups/README.md
+++ b/examples/auto-scaling-groups/README.md
@@ -1,0 +1,45 @@
+# Autscaling Support
+
+This example demonstrates managing a higher level infrastructure
+component - auto scaling groups.
+
+## What
+
+In this example we'll setup:
+
+* A security group
+* A launch configuration
+* An auto scaling group
+* Two scaling policies
+* Two CloudWatch alarms tied to the policies
+
+This provides a fully working autoscaling group setup that will
+autoscale based on increasing or decreasing load on the instances.
+
+## How
+
+With the module installed as described in the README, from this
+directory run:
+
+    puppet apply init.pp --test
+
+
+This should create the AWS resources. You can delete them
+with the `puppet resource` commands when done.
+
+   puppet resource ec2_securitygroup test-sg region=sa-east-1 ensure=absent
+   puppet resource ec2_launchconfiguration test-lc region=sa-east-1 ensure=absent
+   puppet resource ec2_autoscalinggroup test-asg region=sa-east-1 ensure=absent
+   puppet resource ec2_scalingpolicy scaleout region=sa-east-1 ensure=absent
+   puppet resource ec2_scalingpolicy scalein region=sa-east-1 ensure=absent
+   puppet resource cloudwatch_alarm AddCapacity region=sa-east-1 ensure=absent
+   puppet resource cloudwatch_alarm RemoveCapacity region=sa-east-1 ensure=absent
+
+
+## Discussion
+
+The instances managed by the auto scaling group is not directly
+accessible via `puppet resource`. This is by design. However you can
+check on the number of instances currently managed by a group with:
+
+   puppet resource ec2_autoscalinggroup test-asg
--- a/examples/auto-scaling-groups/destroy.pp
+++ b/examples/auto-scaling-groups/destroy.pp
@@ -1,0 +1,21 @@
+cloudwatch_alarm { ['AddCapacity', 'RemoveCapacity']:
+  ensure => absent,
+  region => 'sa-east-1',
+} ~>
+
+ec2_scalingpolicy { ['scaleout', 'scalein']:
+  ensure             => absent,
+  auto_scaling_group => 'test-asg',
+  region             => 'sa-east-1',
+} ~>
+
+ec2_autoscalinggroup { 'test-asg':
+  ensure => absent,
+  region => 'sa-east-1',
+} ~>
+
+ec2_launchconfiguration { 'test-lc':
+  ensure => absent,
+  region => 'sa-east-1',
+}
+
--- a/examples/auto-scaling-groups/init.pp
+++ b/examples/auto-scaling-groups/init.pp
@@ -1,22 +1,22 @@
-ec2_securitygroup { 'garethr-sg-test':
+ec2_securitygroup { 'test-sg':
   ensure      => present,
   description => 'group for testing autoscaling group',
   region      => 'sa-east-1',
 }
 
-ec2_launchconfiguration { 'garethr-lc-test':
+ec2_launchconfiguration { 'test-lc':
   ensure          => present,
-  security_groups => ['garethr-sg-test'],
+  security_groups => ['test-sg'],
   region          => 'sa-east-1',
   image_id        => 'ami-67a60d7a',
   instance_type   => 't1.micro',
 }
 
-ec2_autoscalinggroup { 'gareth-asg-test':
+ec2_autoscalinggroup { 'test-asg':
   ensure               => present,
-  min_size             => 1,
-  max_size             => 3,
+  min_size             => 2,
+  max_size             => 4,
   region               => 'sa-east-1',
-  launch_configuration => 'garethr-lc-test',
-  availability_zones   => ['sa-east-1a'],
+  launch_configuration => 'test-lc',
+  availability_zones   => ['sa-east-1b', 'sa-east-1a'],
 }
--- a/examples/auto-scaling-groups/init.pp
+++ b/examples/auto-scaling-groups/init.pp
@@ -20,3 +20,22 @@ ec2_autoscalinggroup { 'test-asg':
   launch_configuration => 'test-lc',
   availability_zones   => ['sa-east-1b', 'sa-east-1a'],
 }
+
+ec2_scalingpolicy { 'scaleout':
+  ensure             => present,
+  auto_scaling_group => 'test-asg',
+  scaling_adjustment => 30,
+  adjustment_type    => 'PercentChangeInCapacity',
+  region             => 'sa-east-1',
+}
+
+ec2_scalingpolicy { 'scalein':
+  ensure             => present,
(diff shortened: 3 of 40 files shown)


examples/auto-scaling-groups/README.md
```diff
@@ -0,0 +1,47 @@
...
+This should create the AWS resources. You can delete them
+with the `puppet resource` commands when done.
+
+   puppet resource ec2_securitygroup test-sg region=sa-east-1 ensure=absent
+   puppet resource ec2_launchconfiguration test-lc region=sa-east-1 ensure=absent
+   puppet resource ec2_autoscalinggroup test-asg region=sa-east-1 ensure=absent
+   puppet resource ec2_scalingpolicy scaleout region=sa-east-1 ensure=absent
+   puppet resource ec2_scalingpolicy scalein region=sa-east-1 ensure=absent
+   puppet resource cloudwatch_alarm AddCapacity region=sa-east-1 ensure=absent
+   puppet resource cloudwatch_alarm RemoveCapacity region=sa-east-1 ensure=absent
```
cmurphy: These commands don't remove the resources if run in this order:

```bash
$ puppet resource ec2_securitygroup test-sg region=sa-east-1 ensure=absent --modulepath ..
WARNING: Nokogiri was built against LibXML version 2.9.1, but has dynamically loaded 2.9.2
Error: Could not set 'absent' on ensure: There are active instances using security group 'test-sg'
Error: Could not set 'absent' on ensure: There are active instances using security group 'test-sg'
Wrapped exception:
There are active instances using security group 'test-sg'
Error: /Ec2_securitygroup[test-sg]/ensure: change from present to absent failed: Could not set 'absent' on ensure: There are active instances using security group 'test-sg'
ec2_securitygroup { 'test-sg':
  ensure => 'present',
  region => 'sa-east-1',
}
```

As a user testing out these features, I would like to see an example destroy manifest with the resource ordering specified so that I can clean up my proof of concept without worrying about how the resources relate to each other.

lib/puppet/provider/cloudwatch_alarm/v2.rb
```diff
@@ -0,0 +1,106 @@
...
+    :metric=,
+    :namespace=,
+    :statistic=,
+    :period=,
+    :threshold=,
+    :comparison_operator=,
+    :dimensions=,
+    :evaluation_period,
+    :alarm_actions=,
+  ].each{ |method| alias_method method, :update }
```
cmurphy: If I'm reading this correctly, it seems like using `self.flush` would accomplish the same thing and be more idiomatic.

lib/puppet/provider/ec2_scalingpolicy/v2.rb
```diff
@@ -0,0 +1,84 @@
...
+    autoscaling_client(resource[:region]).put_scaling_policy(
+      policy_name: name,
+      auto_scaling_group_name: resource[:auto_scaling_group],
+      scaling_adjustment: resource[:scaling_adjustment],
+      adjustment_type: resource[:adjustment_type],
+    )
+  end
+
+  def auto_scaling_group=(value)
+    update
```
cmurphy: Calling the method directly instead of aliasing seems inconsistent with how it's done in the cloudwatch provider, and again this could be done with the `flush` method.

lib/puppet/type/cloudwatch_alarm.rb
```diff
@@ -0,0 +1,91 @@
+Puppet::Type.newtype(:cloudwatch_alarm) do
+  @doc = 'type representing an AWS CloudWatch Alarm'
```
cmurphy: Any reason to use `@doc` instead of the `desc` method?

lib/puppet/type/cloudwatch_alarm.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  ensurable
+
+  newparam(:name, namevar: true) do
+    desc 'the name of the alarm'
```
cmurphy: Sort of a nitpick, but I'd like to see the first letter capitalized and the phrase ended with a period. The `ensure` property has pregenerated documentation that uses that style, so it's more pleasant to see consistent style when running `puppet describe`:

```bash
$ puppet describe cloudwatch_alarm --modulepath ..

cloudwatch_alarm
================
type representing an AWS CloudWatch Alarm


Parameters
----------

- **alarm_actions**
    the actions to trigger when the alarm triggers

- **comparison_operator**
    the operator to use to test the metric

- **dimensions**
    the dimensions to filter the alerm by

- **ensure**
    The basic property that the resource should be in.
    Valid values are `present`, `absent`.

- **evaluation_periods**
    the number of checks to use to confirm the alarm
...
```

spec/unit/type/cloudwatch_alarm_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+      :statistic,
+      :period,
+    ]
+  end
+
+  it 'should have expected properties' do
+    properties.each do |property|
+      expect(type_class.properties.map(&:name)).to be_include(property)
+    end
+  end
```
cmurphy: This is not really a useful test, in my opinion. It is much more important to test that the parameter validation works. Example: https://github.com/puppetlabs/puppetlabs-mysql/blob/master/spec/unit/puppet/type/mysql_user_spec.rb

lib/puppet/type/cloudwatch_alarm.rb
```diff
@@ -0,0 +1,91 @@
...
+      fail 'statistic must not be blank' if value == ''
+    end
+  end
+
+  newproperty(:period) do
+    desc 'the periodicity of the alarm check'
+    validate do |value|
+      fail 'period cannot be blank' if value == ''
+    end
+    def insync?(is)
```
cmurphy: Why override insync? instead of munging? Does the `is` value typically return a string?

cmurphy:
Any particular reason there's no unit tests for the providers?

lib/puppet/type/cloudwatch_alarm.rb
garethr: Nope, is one preferred over the other?

spec/unit/type/cloudwatch_alarm_spec.rb
garethr: Expanded the type tests in a future commit: https://github.com/garethr/puppetlabs-aws/commit/fe92a4ecb8846b17ced69850fdaac0c01a156690

lib/puppet/type/cloudwatch_alarm.rb
garethr: Thanks, fixed for the autoscaling group types in a future commit. I'll change the other types as well separately to avoid merge issues.

lib/puppet/provider/ec2_scalingpolicy/v2.rb
garethr: Thanks, changed in future commit.

examples/auto-scaling-groups/README.md
garethr: Thanks, changed in future commit.

lib/puppet/provider/cloudwatch_alarm/v2.rb
garethr: Agreed, changed to flush in future commit.

examples/auto-scaling-groups/README.md
```diff
@@ -0,0 +1,47 @@
+# Autscaling Support
```
garethr: Fixed, thanks.

garethr:
@cmurphy thanks for the view, I think I've addressed all the points raised in the commits this morning. Any other points let me know. All much appreciated. Some of this I'll back port to existing stuff as well.

lib/puppet/type/cloudwatch_alarm.rb
```diff
@@ -0,0 +1,91 @@
...
+  newproperty(:region) do
+    desc 'The region in which to launch the instances.'
+    validate do |value|
+      fail 'region should not contain spaces' if value =~ /\s/
+      fail 'region should not be blank' if value == ''
+    end
+  end
+
+  newparam(:dimensions, :array_matching => :all) do
+    desc 'The dimensions to filter the alerm by.'
```
cmurphy: Typo here: alerm/alarm

cmurphy: Should these parameters have validation as well? It looks like https://github.com/puppetlabs/puppetlabs-aws/pull/30/files#diff-84de254bad7394aa4014979a5b0fde38R70 requires this this to be a hash, the example uses an array, and it (predictably) messes things up when I try to use a string.

cmurphy:
All my points seem to have been addressed, but on second review I discovered a couple of minor things.

examples/auto-scaling-groups/README.md
```diff
@@ -0,0 +1,47 @@
...
+## How
+
+With the module installed as described in the README, from this
+directory run:
+
+    puppet apply init.pp --test
+
+
+This should create the AWS resources. You can delete them
+with the the destory manifest when done.
```
cmurphy: Typo here, destory/destroy

examples/auto-scaling-groups/README.md
```diff
@@ -0,0 +1,47 @@
...
+
+    puppet apply destroy.pp --test
+
+Note that this will shut down any instances launched by the auto scaling
+grup. Once the instances are terminated you can remove the accompanying
```
cmurphy: Typo, grup/group

examples/auto-scaling-groups/init.pp
```diff
@@ -0,0 +1,71 @@
...
+  region               => 'sa-east-1',
+  launch_configuration => 'test-lc',
+  availability_zones   => ['sa-east-1b', 'sa-east-1a'],
+}
+
+ec2_scalingpolicy { 'scaleout':
+  ensure             => present,
+  auto_scaling_group => 'test-asg',
+  scaling_adjustment => 30,
+  adjustment_type    => 'PercentChangeInCapacity',
```
cmurphy: After going through the tutorial, it's not intuitive to me where this keyword comes from and what values could be put in here. I'd like to see this documented in the type parameter and perhaps validated with newvalues() if there are a limited number of allowed keywords here. Same with ChangeInCapacity in the scalin policy and the metric, statistic, comparison_operator, in the alarm resources.

lib/puppet/type/cloudwatch_alarm.rb
cmurphy: I like using `desc` to be consistent with how the types are documented, otherwise I'm not sure of the difference.

spec/acceptance/autoscaling_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    it 'should create CloudWatch alarms' do
+      find_alarm("#{@name}-AddCapacity")
+      find_alarm("#{@name}-RemoveCapacity")
+    end
+
+    it 'should create scaling policies' do
+      find_scaling_policy("#{@name}-scaleout", "#{@name}-asg")
+      find_scaling_policy("#{@name}-scalein", "#{@name}-asg")
+    end
+
```
cmurphy: It looks like this is only testing resource creation. The cleanup part isn't validated, it's just assumed to work. Would it also be useful to have 'it' assertions on the destruction, update, and idempotency of resources?

cmurphy:
Had a few more notes after chatting with Ethan, going through the autoscaling tutorial, and looking through the code again.

examples/auto-scaling-groups/init.pp
garethr: I decided against adding validation to the types for a number of the properties for a few reasons (that are worth discussing here).

* The API is large, and evolving. So new options could be added by Amazon at any time. If we embed validation in the types we have to monitor that and release new versions. Not including validation minimises the maintenance overhead.
* The error messages from the API client are in my experience very good at doing the validation. For instance if you run this manifest with a random value you get the following:

```
Error: Could not set 'present' on ensure: Valid adjustment types are: [PercentChangeInCapacity, ChangeInCapacity, ExactCapacity]. at 30:/Users/garethr/Documents/puppetlabs-aws/tests/asg.pp
```

* It raises the barrier to entry for contributions from third parties, in some cases the additional code for validation rules and testing those rules would be greater than the provider adding the new functionality. The size of the API means I'd favour as low a barrier to entry as possible.

With regards to what the options mean, the module should require a solid understanding of the AWS primitives. It's not intended to be an abstraction (like say the Nginx module) but rather an alternative user interface. The example currently links to http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/WhatIsAutoScaling.html which covers the concepts.

spec/acceptance/autoscaling_spec.rb
garethr: I've added a bunch more tests here to assert on various properties, and to change those properties.

I've not added deletion tests as the waiters aren't implemented for the autoscaling resources yet.  We can add our own implementation if we thing it's worth it, but I'd probably rather wait and add them later with the native wait.

lib/puppet/type/cloudwatch_alarm.rb
garethr: A few other types I'd checked use @doc eg:https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/cron.rb, https://github.com/puppetlabs/puppet/blob/master/lib/puppet/type/augeas.rb

cmurphy:
:+1: code looks good to me. The only other thing is I'd prefer to see the commits rebased into just a few logical units.

Iristyle:
:+1: to @cmurphy comment about squashing commits where appropriate.  I don't think we need to keep commits with `fix typos` or that remove code / comments that were temporary, given this hasn't hit master branch yet.  It just adds noise to the history.

If you could also mention the resource type in the commit subject where it's missing that would be good - some of the messages could really apply to any types, and it will help when we need to look at the history later.

examples/auto-scaling-groups/destroy.pp
```diff
@@ -0,0 +1,21 @@
+cloudwatch_alarm { ['AddCapacity', 'RemoveCapacity']:
+  ensure => absent,
+  region => 'sa-east-1',
+} ~>
+
+ec2_scalingpolicy { ['scaleout', 'scalein']:
```
Iristyle: I'm not so sure about the `ec2` prefixes for ASG resources... given these are not part of the EC2 API, and we don't use the EC2 client to configure them, but rather the ASG client instead.

garethr: That rationale here was that Autoscaling is part of the EC2 (customer facing) service, rather than a separate service. For instance the AWS dashboard lists Autoscaling as part of the EC2 options.

My gut feeling here was that going with the user facing name, rather than the API implementation, was the better bet.

examples/auto-scaling-groups/init.pp
Iristyle: Normally I would agree here @cmurphy, but @garethr is correct about reduced maint burden.  The AWS service endpoints perform the validation for us.

This is not always the most user-friendly option given it requires an extra network roundtrip, but the size of the API dictates that it's going to be impractically difficult for us to validate "all the things".

If you take a look at the Ruby SDK code, you'll see that they've chosen the same server-side only validation approach (and only have basic schema definitions around the various properties like `AdjustmentType`):
https://github.com/aws/aws-sdk-core-ruby/blob/c51717d4fc68cf7729876ef73b6622b8b51d4c09/aws-sdk-core/apis/AutoScaling.api.json#L786-L791

examples/auto-scaling-groups/destroy.pp
Iristyle: That's a fair enough point... I can be swayed in that direction.

garethr:
@Iristyle I've rebased this to squash and fixup a bunch of commits.

PR #35, opened 2014-11-27
Add route53 record support

garethr:
Not ready for merging just yet, but just to get visibility onto this in the short term.

* This currently supports managing zones, and txt, a and ns records
* I'd in particular be interested in feedback on the approach to the record types

### DIFF
--- a/examples/manage-dns/README.md
+++ b/examples/manage-dns/README.md
@@ -1,0 +1,72 @@
+# DNS
+
+This example demonstrates using the Route53 resources to manage DNS
+records.
+
+## What
+
+For this example we'll create a Zone, along with an A record. We'll also
+show how to list current records of the supported types.
+
+The Zone we're creating is for `puppetlabs.com`, although when you run
+the example you may want to change that to your own domain.
+
+## How
+
+With the module installed as described in the README, from this
+directory run:
+
+    puppet apply init.pp --test
+
+
+This should create a ROute53 zone and an A record for
+`local.puppetlabs.com` pointing to 127.0.0.1.
+
+You should be see the zone when you run:
+
+    puppet resource route53_zone
+
+This should output the following, maybe along with other domains you're
+already managing.
+
+```puppet
+route53_zone { 'puppetlabs.com.':
+  ensure => 'present',
+}
+```
+
+And to list the A records:
+
+    puppet resource route53_a_record
+
+Which should return:
+
+```puppet
+route53_a_record { 'local.puppetlabs.com.':
+  ensure => 'present',
+  ttl    => '3000',
+  values => ['127.0.0.1'],
+  zone   => 'puppetlabs.com.',
+}
+```
+
+We can also list the automatically created NS records with:
+
+    puppet resource route53_ns_record
+
+Which should output something like:
+
+```puppet
+route53_ns_record { 'puppetlabs.com.':
+  ensure => 'present',
+  ttl    => '172800',
+  values => ['ns-1254.awsdns-28.org.', 'ns-1922.awsdns-48.co.uk.', 'ns-806.awsdns-36.net.', 'ns-103.awsdns-12.com.'],
+  zone   => 'puppetlabs.com.',
+}
+```
+
+Finally we can delete the A record followed by the Zone:
+
+    puppet resource route53_a_record local.puppetlabs.com. zone=puppetlabs.com. values=127.0.0.1 ttl=3000 ensure=absent
+    puppet resource route53_zone puppetlabs.com. ensure=absent
+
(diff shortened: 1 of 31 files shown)


examples/manage-dns/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## How
+
+With the module installed as described in the README, from this
+directory run:
+
+    puppet apply init.pp --test
+
+
+This should create a ROute53 zone and an A record for
```
cmurphy: Typo here, ROute53 vs Route53

cmurphy: This also creates an NS record and a TXT record

examples/manage-dns/init.pp
```diff
@@ -0,0 +1,28 @@
...
+    'ns-1724.awsdns-23.co.uk.',
+    'ns-969.awsdns-57.net.',
+  ]
+}
+
+route53_txt_record { 'local.puppetlabs.com.':
+  ensure => present,
+  zone   => 'puppetlabs.com.',
+  ttl    => 17200,
+  values => '"message"'
```
cmurphy: The reasoning for why this is `'"message"'` instead of `'message'` is unclear, and the error message when not including the double quotes is not that helpful:

```
Error: /Stage[main]/Main/Route53_txt_record[local.test.plabs.net.]: Could not evaluate: Invalid Resource Record: FATAL problem: InvalidTXTRDATA encountered at message
```

Maybe some extra validation or munging would be helpful here?

lib/puppet/provider/route53_record.rb
```diff
@@ -0,0 +1,83 @@
...
+            ttl: resource[:ttl],
+            resource_records: records,
+          }
+        }]
+      }
+    }
+  end
+
+  def zone_id
+    zones = route53_client.list_hosted_zones.data.hosted_zones.select { |zone| zone.name == resource[:zone] }
```
cmurphy: This line is a bit long, splitting it into a do/end block would be a little nicer.

examples/manage-dns/README.md
```diff
@@ -0,0 +1,72 @@
...
+  ensure => 'present',
+  ttl    => '172800',
+  values => ['ns-1254.awsdns-28.org.', 'ns-1922.awsdns-48.co.uk.', 'ns-806.awsdns-36.net.', 'ns-103.awsdns-12.com.'],
+  zone   => 'puppetlabs.com.',
+}
+```
+
+Finally we can delete the A record followed by the Zone:
+
+    puppet resource route53_a_record local.puppetlabs.com. zone=puppetlabs.com. values=127.0.0.1 ttl=3000 ensure=absent
```
cmurphy: This forgets the TXT and NS records that were created.

The zone seems to create an NS record by default with the same name as the zone, and the example creates an additional NS record with a 'local' subdomain. One of the NS records needs to be deleted before the zone can be deleted.

A destroy.pp manifest would be nice here.

It's kind of annoying to have to specify all the parameters in the `puppet resource` command in order to delete a resource, especially if one of the parameters is `values => ['ns-1254.awsdns-28.org.', 'ns-1922.awsdns-48.co.uk.', 'ns-806.awsdns-36.net.', 'ns-103.awsdns-12.com.',`. With other resources (I think) we rely on the name being unique, is that not sufficient here?

examples/manage-dns/init.pp
cmurphy: I understand that the double quotes are needed when entering it in the console, and this is consistent with that, but from the puppet side it's not really clear.

examples/manage-dns/README.md
garethr: Fixed typo and updated example readme to match the records being created.

lib/puppet/provider/route53_record.rb
garethr: Agreed, split.

examples/manage-dns/README.md
garethr: I've added a destroy manifest which cleans up. The automatically created SOA and NS records are attached the zone behind the scenes. See the delete_hosted_zone docs for the Route53 Client on http://docs.aws.amazon.com/sdkforruby/api/frames.html.

You always have to send all the information to make changes to the records, but I've just swapped this over to populate the details from the @property_hash ifyou don'e provide them.

You should now be able to just run:

    puppet resource route53_txt_record local.puppetlabs.com. ensure=absent

For example and it should just work. Thanks.

examples/manage-dns/init.pp
garethr: Good point. This makes sense from a DNS point of view but it's not as nice from a pure Puppet point of view. I've added in munging if the quotes are not added. I've also changed this in the example.

garethr:
@cmurphy Thanks, I've responded to and addressed the points you raised.

cmurphy:
Looks good, but could you rebase the 25 commits into just a few?

PR #43, opened 2015-01-06
Add VPC support

garethr:
This is intended to replace https://github.com/puppetlabs/puppetlabs-aws/pull/19. It's based on the work by Yelp which covers a really good real world VPC use case.

I'm working through adding providers using the v2 SDK.


Ziaunys:
+1 Excited to have VPC support!

lib/puppet/provider/ec2_vpc/v2.rb
```diff
@@ -0,0 +1,94 @@
+require_relative '../../../puppet_x/puppetlabs/aws.rb'
+
+Puppet::Type.type(:ec2_vpc).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) do
+  confine feature: :aws
+
+  mk_resource_methods
+  remove_method :tags=
```
cmurphy: Why is this here? Shouldn't users be able to set tags? If not, should these be a read-only property, or not a property?

lib/puppet/provider/ec2_vpc/v2.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def self.instances
+    regions.collect do |region|
+      response = ec2_client(region).describe_vpcs()
+      vpcs = []
+        response.data.vpcs.each do |vpc|
```
cmurphy: Indendation is kind of weird here, I think this line is meant to be aligned with the line above, with the next line indented.

lib/puppet/provider/ec2_vpc_routetable/v2.rb
```diff
@@ -0,0 +1,113 @@
...
+    else
+      begin
+        igw_response = ec2.describe_internet_gateways(internet_gateway_ids: [route.gateway_id])
+        name_from_tag(igw_response.data.internet_gateways.first)
+      rescue Aws::EC2::Errors::InvalidInternetGatewayIDNotFound
+        nil
+      end
+    end
+    {
+      'destination_cidr_block' => route.destination_cidr_block,
```
cmurphy: Using a string instead of a symbol seems inconsistent with the other *_to_hash methods, is there a reason for that?

lib/puppet/provider/ec2_vpc_vpn_gateway/v2.rb
```diff
@@ -0,0 +1,111 @@
...
+  def destroy
+    region = @property_hash[:region]
+    Puppet.info("Destroying VPN gateway #{name} in #{region}")
+    ec2 = ec2_client(region)
+    vpc_id = @property_hash[:vpc_id]
+    ec2.detach_vpn_gateway(
+      vpn_gateway_id: @property_hash[:id],
+      vpc_id: vpc_id,
+    ) if vpc_id
+    sleep 60 # TODO should be able to wait for detachment
```
cmurphy: Should we complete this TODO before merging?

lib/puppet/type/ec2_vpc_vpn.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  newproperty(:customer_gateway) do
+    desc 'The customer gateway to attach to the VPN.'
+  end
+
+  newproperty(:type) do
+    desc 'The type of VPN gateway, defaults to ipsec.1.'
+    defaultto 'ipsec.1'
+    validate do |value|
+      unless value =~ /^ipsec\.1$/
```
cmurphy: Why is this a writeable property if there is only one valid value? This should either not be a property or be a read-only property to reflect how AWS reports it.

spec/acceptance/vpc_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    after(:all) do
+      new_config = @config.update({:ensure => 'absent'})
+      template = 'vpc_delete.pp.tmpl'
+      PuppetManifest.new(template, new_config).apply
+    end
+
+    it 'should run successfully with changes' do
+      expect(@exit.exitstatus).to eq(2)
+    end
```
cmurphy: I'd like to see another test for idempotency here.

tests/vpc.pp
```diff
@@ -0,0 +1,62 @@
+ec2_vpc { 'garethr-test':
```
cmurphy: Maybe pick a more generalized descriptor for these.

cmurphy:
For other features we had a directory in examples/ with a README, an init.pp, and a delete.pp. I think it would be nice to see that here too, instead of just an unexplained manifest in the tests/ directory.

lib/puppet/provider/ec2_vpc/v2.rb
garethr: If you run the code you should find you can modify tags.

It's there because `mk_resource_methods` creates a `tags=` method on the new class. But the parent class has the implementation.

`remove_method`: "Removes the method identified by symbol from the current class. For an example, see Module.undef_method."

So this removes the method added by `mk_resource_methods` and uses the one in the base class instead.

lib/puppet/provider/ec2_vpc_vpn_gateway/v2.rb
garethr: Currently the SDK doesn't yet have a waiter for detached vpn gateways. This basically replicated the wait used in the Yelp provider if memory serves.

lib/puppet/type/ec2_vpc_vpn.rb
garethr: I've replicated the API here quite closely (http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVpnConnection.html), which allows you to pass a value here but currently only allows the single value. This to me suggested that might change at some point in the future, so allowing this to be explicit made sense.

tests/vpc.pp
garethr: I'll drop this and convert into a simple example with documentation.

lib/puppet/provider/ec2_vpc_routetable/v2.rb
garethr: Unfortunately yes. See the same thing in the security group provider too.

For nested values (like this and the ingress rules) the comparison of the array within insync relies on these being strings, as Puppet automatically coerces the other side of the equality to strings.

spec/acceptance/vpc_spec.rb
garethr: I've added a whole bunch more acceptance tests and assertions, including this check, in a future commit.

lib/puppet/provider/ec2_vpc/v2.rb
garethr: Resolved, thanks.

lib/puppet/provider/ec2_vpc/v2.rb
garethr: Just a note that in extending the test suite I've added tests demonstrating the tagging of resources.

garethr:
@cmurphy thanks for comments, I've replied to each individually and have just converted the test manifest into a simple example. Any more feedback let me know.

lib/puppet/provider/ec2_vpc_vpn_gateway/v2.rb
cmurphy: 60 unconditional seconds seems like an eternity to be blocked on one provider. Could this instead poll the API until it's removed or it times out?

examples/vpc-example/init.pp
```diff
@@ -0,0 +1,41 @@
...
+  ensure       => present,
+  region       => 'sa-east-1',
+  cidr_block   => '10.0.0.0/16',
+  #  dhcp_options => 'sample-options',
+}
+
+#ec2_vpc_dhcp_options { 'sample-options':
+#  ensure => present,
+#  region => 'sa-east-1',
+#}
```
cmurphy: Why are these things commented out?

examples/vpc-example/README.md
```diff
@@ -0,0 +1,59 @@
...
+parts of VPC in the Puppet DSL. This can be useful for creating and
+sharing duplicate environments or alternatively as part of an audit
+mechanism to ensure your VPC configuration is as intended.
+
+## What
+
+For this examplei= we're going to create a VPC, a subnet, a route table
+and an internet gateway. This broadly follows the [getting started
+example](http://docs.aws.amazon.com/AmazonVPC/latest/GettingStartedGuide/Wizard.html)
+from the official Amazon documentation.
```
cmurphy: It would be nice to see the rest of the providers demonstrated too, the way the previous tests/vpc.pp did.

examples/vpc-example/README.md
```diff
@@ -0,0 +1,59 @@
...
+A virtual private cloud (VPC) is a virtual network that closely
+resembles a traditional network that you'd operate in your own data
+cente. The AWS Puppet module allows for describing the various moving
+parts of VPC in the Puppet DSL. This can be useful for creating and
+sharing duplicate environments or alternatively as part of an audit
+mechanism to ensure your VPC configuration is as intended.
+
+## What
+
+For this examplei= we're going to create a VPC, a subnet, a route table
```
cmurphy: Typo - "examplei="

examples/vpc-example/README.md
```diff
@@ -0,0 +1,59 @@
...
+## How
+
+With the module installed as described in the README, from this
+directory run:
+
+    puppet apply init.pp --test
+
+This should create the resources discussed above in the AWS example.
+Once complete you should be able to observe the VPC and associated
+resources via the AWS console as shown in the example. Alternatovely you
```
cmurphy: Typo - "Alternatovely"

examples/vpc-example/README.md
```diff
@@ -0,0 +1,59 @@
+# AWS VPC
+
+A virtual private cloud (VPC) is a virtual network that closely
+resembles a traditional network that you'd operate in your own data
+cente. The AWS Puppet module allows for describing the various moving
```
cmurphy: Typo, "cente"

lib/puppet/provider/ec2_vpc_vpn_gateway/v2.rb
garethr: I'll change this to retry on the relevant exception, using the same approach for tagging instances. Unfortunately the detachment does appear to generally take a while.

examples/vpc-example/README.md
garethr: I'll create a separate example for the VPN resources. The issue with those is that you can't really having a 'working' example without having the hardware at your end - we can just create some resources in AWS with dummy data.

garethr:
@cmurphy I've replaced the sleep with a retry as discussed, and I've extended the VPC example to include a separate VPN example with the other resources. Let me know if this needs anything else.

cmurphy:
+1 looks good. As usual, could you rebase away some of the fix commits into the larger feature commits?

garethr:
@cmurphy Cheers, rebased to reduce the number of commits.

PR #44, opened 2015-01-09
First integration test commit

zreichert:
This commit sets up the environment needed to perform end to end integration test with Beaker-Rspec.
Currently this will not tear down correctly. A bug was found that prevents the after(:all) step from being executed correctly.  https://tickets.puppetlabs.com/browse/CLOUD-184

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -27,6 +27,16 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
     t.pattern = 'spec/acceptance'
 end
 
+desc "Run integration tests"
+RSpec::Core::RakeTask.new(:integration) do |t|
+  t.pattern = 'spec/acceptance/integration'
+  ENV['BEAKER_setfile'] ||= 'spec/acceptance/beaker/nodesets/rhel7.yaml'
+  ENV['SPEC_FORGE'] ||= 'https://api-forge-aio01-qatest.puppetlabs.com/'
+  unless ENV['SPEC_VERSION']
+    raise StandardError 'You must specify the environment variable SPEC_VERSION'
+  end
+end
+
 task :metadata do
   sh "metadata-json-lint metadata.json"
 end
--- a/spec/beaker_configure.rb
+++ b/spec/beaker_configure.rb
@@ -1,0 +1,48 @@
+require 'beaker-rspec'
+
+install_pe
+
+RSpec.configure do |c|
+
+  c.before :suite do
+    # find the agent that is the provisioner
+    @provisioner = find_only_one(:provisioner)
+    #read the AWS credential from Jenkins slave
+    home = ENV['HOME']
+    file = File.open("#{home}/.aws/credentials")
+    # configure the agent
+    agent_home = on(@provisioner, 'printenv HOME').stdout.chomp
+    on(@provisioner, puppet("module install puppetlabs-pe_gem"))
+    pp =<<-EOS
+    package{'aws-sdk-core':
+      ensure    => present,
+      provider  => pe_gem,
+    }
+    EOS
+    apply_manifest_on(@provisioner, pp)
+    on(@provisioner, "mkdir #{agent_home}/.aws")
+    create_remote_file(@provisioner, "#{agent_home}/.aws/credentials", file.read)
+    # configure the master
+    on(master, '/opt/puppet/bin/puppetserver gem install aws-sdk-core')
+    # restart puppet server
+    on(master, "puppet resource service pe-puppetserver ensure=stopped")
+    on(master, "puppet resource service pe-puppetserver ensure=running")
+    masterHostName = on(master, "hostname").stdout.chomp
+    i = 0
+    # -k to ignore HTTPS error that isn't relevant to us
+    curl_call = "-I -k https://#{masterHostName}:8140/production/certificate_statuses/all"
+    while i < 35 do
+      sleep 5
+      i += 1
+      exit_code = curl_on(master, curl_call, :acceptable_exit_codes => [0,1,7]).exit_code
+      # Exit code 7 is "connection refused"
+      if exit_code != '7'
+        sleep 20
+        puts 'Restarting the Puppet Server was successful!'
+        break
+      end
+    end
+
+  end
+
+end
--- a/spec/integration/README.md
+++ b/spec/integration/README.md
@@ -1,0 +1,16 @@
+Integration Tests
+========================
+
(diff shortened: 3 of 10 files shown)


spec/integration/integration_1_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      create_remote_file(master, '/etc/puppetlabs/puppet/environments/production/manifests/site.pp', @manifest)
+      on(master, 'chmod 777 /etc/puppetlabs/puppet/environments/production/manifests/site.pp')
+      # initiate a puppet run
+      on(@provisioner, puppet('agent --test'), {:acceptable_exit_codes => [0,2]})
+    end
+
+    after(:all)
+      @instance = @aws.get_instance(@config[:instance_name])
+      @aws.ec2_client.wait_until(:instance_running, instance_ids:[@instance.instance_id])
+      on(@provisioner, "puppet resource ec2_instance #{@instance_name} ensure=absent")
```
garethr: ah, you have to specify the region in order to scope this. Basically names are region unique, so this also needs region=sa-east-1

Rakefile
```diff
@@ -27,6 +27,11 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
     t.pattern = 'spec/acceptance'
 end
 
+desc "Run integration tests"
+task :integration do
+  sh("BEAKER_setfile=spec/acceptance/beaker/nodesets/rhel7.yaml bundle exec rspec spec/acceptance/integration")
```
garethr: This would be better to use the built-in rake task so as to allow overriding the individual specs and the set as needed using the standard environment variables. ie.

```ruby
RSpec::Core::RakeTask.new(:acceptance) do |t|
  t.pattern = 'spec/acceptance/integration'
end
```

spec/integration/README.md
```diff
@@ -0,0 +1,16 @@
...
+This folder contains integration tests for the puppetlabs-aws project. These tests were originally written by the QA team at Puppet Labs and is actively maintained by the QA team. Feel free to contribute tests to this folder as long as they are written with [Beaker-RSpec](https://github.com/puppetlabs/beaker-rspec) and follow the guidelines below.
+
+## Integration?
+
+The puppetlabs-aws project already contains RSpec tests and you might be wondering why there is a need to have a set of tests separate from those tests. At Puppet Labs we define an "integration" test as:
+
+>Validating the system state and/or side effects while completing a complete life cycle of user stories using a system. This type of test crosses the boundary of a discrete tool in the process of testing a defined user objective that utilizes a system composed of integrated components.  What this means for this project is that we will install and configure all infrastructure used in a real-world PE environment.
+
+## Integration Test Requirements
+These tests are internal to puppet labs.  The following is a list of requirements to run these tests.
```
garethr: I don't see a reason for these tests to be specific to Puppet Labs given this is an open source project.

spec/spec_helper_acceptance.rb
```diff
@@ -93,6 +94,15 @@ def get_instances(name)
     end.flatten
   end
 
+  def get_instance(name)
```
garethr: these methods duplicate those currently found in the individual tests. ie. https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/acceptance/instance_spec.rb#L13-L17.

Happy for this to move here, but we should change the other specs as well.

garethr:
@zreichert thanks for getting this up for review. Given how close the implementation of this test is to the existing tests, it would in my view be best to take a different approach - namely that we could run the all the EC2 accessing tests either locally (ie. what we already have) or remotely (ie. what you propose here, using a master and beaker).

I think if we have acceptance tests and integration tests that are basically the same thing apart from how they are run we're just going to confuse, plus we'll repeat work in multiple places or have areas that are only tested in one.

From a quick look this should just involve abstracting the actual puppet run, plus providing a toggle between the two mechanisms (basically the spec helper). Happy to pair on that if we can find a suitable time.

spec/integration/nodesets/rhel7.yaml
```diff
@@ -0,0 +1,24 @@
+HOSTS:
+  master:
+    roles:
+      - master
+      - agent
+      - database
+      - dashboard
+    platform: el-7-x86_64
+    hypervisor: vcloud
```
garethr: It would be useful to include a nodeset which can run locally as well using vagrant that can be toggled using the ENV variables.

spec/integration/README.md
Iristyle: Agreed

Rakefile
```diff
@@ -27,6 +27,16 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
     t.pattern = 'spec/acceptance'
 end
 
+desc "Run integration tests"
+RSpec::Core::RakeTask.new(:integration) do |t|
```
Iristyle: The `Gemfile` has `:system_tests` and here we have `:integration`.  Would be good to pick one nomenclature.

spec/beaker_configure.rb
```diff
@@ -0,0 +1,48 @@
...
+    # configure the master
+    on(master, '/opt/puppet/bin/puppetserver gem install aws-sdk-core')
+    # restart puppet server
+    on(master, "puppet resource service pe-puppetserver ensure=stopped")
+    on(master, "puppet resource service pe-puppetserver ensure=running")
+    masterHostName = on(master, "hostname").stdout.chomp
+    i = 0
+    # -k to ignore HTTPS error that isn't relevant to us
+    curl_call = "-I -k https://#{masterHostName}:8140/production/certificate_statuses/all"
+    while i < 100 do
```
Iristyle: Sleeping a max of 2500 seconds seems like overkill, no?

spec/integration/integration_1_spec.rb
```diff
@@ -0,0 +1,95 @@
+require 'spec_helper_acceptance'
+require 'securerandom'
+require 'beaker_configure'
+
+describe 'User Scenario instaling via PMT master & agent' do
```
Iristyle: nit: `installing`

Rakefile
zreichert: Removed :system_tests in favor of :integration

justinstoller: Some modules already have a gem group called "system_tests". We should either make this project match them, or the other modules match this one.

Rakefile
```diff
@@ -27,6 +27,16 @@ RSpec::Core::RakeTask.new(:acceptance) do |t|
...
+  t.pattern = 'spec/acceptance/integration'
+  ENV['BEAKER_setfile'] ||= 'spec/acceptance/beaker/nodesets/rhel7.yaml'
+  ENV['SPEC_FORGE'] ||= 'https://api-forge-aio01-qatest.puppetlabs.com/'
```
justinstoller: This should be a service name, not the actually host name (modules-staging.puppetlabs.com)

Iristyle:
Current status on this PR @zreichert ?

Thanks!

PR #45, opened 2015-01-14
Add an example of using the Puppet testing tools with AWS manifests

garethr:
I think one of the compelling reasons for using Puppet for describing infrastructure is the existing tooling already available. This simple demo shows how to using rspec-puppet to write unit tests against AWS manifests - this would make a great CD pipeline for changes to AWS.


garethr:
Merging as just a docs update.

PR #47, opened 2015-01-15
Use a flag to run tests under beaker

garethr:
Rather than have two parallel test suites, this proposal allows toggling
tests to be run using beaker with an ENV variable.

This is currently a working demonstration with a number of hard coded
bits, in particular:

* Uses Puppet rather than allowing for PE as well
* Uses a single node as the master and agent
* Uses Vagrant as the backend, rather than allowing for Vagrant and vmspooler

These are all easily changeable, but want to get feedback on the
approach first.

Note that this is based on #44, but takes a different approach. This has been tested using the existing unchanged load banacer spec using:

```
VAGRANT_DEFAULT_PROVIDER=virtualbox SPEC=spec/acceptance/loadbalancer_spec.rb BEAKER_provision=yes BEAKER_destroy=no USE_BEAKER=true be rake spec
```

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -2,6 +2,9 @@ require 'aws-sdk-core'
 require 'mustache'
 require 'open3'
 
+require 'spec_helper_integration' if ENV['USE_BEAKER']
+
+
 class PuppetManifest < Mustache
 
   def initialize(file, config)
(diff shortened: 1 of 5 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -14,6 +17,17 @@ def initialize(file, config)
   end
 
   def apply
+    ENV['USE_BEAKER'] ||= false
```
justinstoller: you can't set ENV variables to booleans. It must be "false".

spec/spec_helper_acceptance.rb
```diff
@@ -2,6 +2,9 @@
 require 'mustache'
 require 'open3'
 
+require 'spec_helper_integration' if ENV['USE_BEAKER']
```
justinstoller: I wouldn't call this "spec_helper_integration" as that has connotations that it will be loaded directly by tests. I would set it as a regular file loaded from the support dir.

Also, if a user explicitly sets `USE_BEAKER` to false on the shell, this will be "false" (String) and still be "truthy"

garethr:
@zreichert I'll close this for now as more a what's possible rather than a merge now and will see what you come up with.

PR #51, opened 2015-01-18
Use SPDX string for license as suggested by Forge

ghoneycutt:
http://spdx.org/licenses/

License is still Apache Software License v2


garethr:
@ghoneycutt thanks, much appreciated.

PR #52, opened 2015-01-19
(CLOUD-142, CLOUD-202) Better handling of stopping instances

garethr:
This started out as just making the log messages nicer, so you now get:

* changed running to stopped
* changed stopped to running

I also improved other notice messages or their placement around stopped,
and better handled using running as an alias of present. Ensure for
puppet resources will also now correctly report the status rather than
just present or absent.

Also a minor fix for a typo affecting the output of information in puppet resource.


Iristyle:
Some of the tests that should be in 0b0e6fa ended up in #53 accidentally

Iristyle:
You can hack 1c0099b out of here... I cherry-picked it over to master since that was an obvious thing to include.

garethr:
Thanks, I've pulled that commit out and moved those tests into the relevant commits.

cmurphy:
:+1:

Iristyle:
Validated locally - merging. :+1:

PR #53, opened 2015-01-19
(CLOUD-143) Improve formatting of instance tags when output in logs

garethr:
Previously tags would be output in the order they were entered or
returned from the AWS API, which might be different. The string
representations could also use single quotes or double quotes. All of
which made parsing changes by eye difficult. This change normalises both
sides of the is and should to be the same, as well as ordering by key
name.

### DIFF
--- a/lib/puppet/type/ec2_instance.rb
+++ b/lib/puppet/type/ec2_instance.rb
@@ -1,3 +1,5 @@
+require_relative '../../puppet_x/puppetlabs/property/tag.rb'
+
 Puppet::Type.newtype(:ec2_instance) do
   @doc = 'type representing an EC2 instance'
 
@@ -37,7 +39,7 @@ Puppet::Type.newtype(:ec2_instance) do
     desc 'the security groups to associate the instance'
   end
 
-  newproperty(:tags) do
+  newproperty(:tags, :parent => PuppetX::Property::AwsTag) do
     desc 'the tags for the instance'
   end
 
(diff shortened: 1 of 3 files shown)


Iristyle:
FYI - on platform side, we usually annotate commits with the ticket number in the subject line, and while it probably isn't necessary for our v1 release, probably a good habit to get into, as it will affect incoming bug reports against future releases, etc.

`(CLOUD-143) Message goes here`

Iristyle:
This looks pretty good except for the unrelated tests that made their way into c151e70

garethr:
Thanks, pulled out the tests into the relevant PR.

cmurphy:
+1 but needs a rebase.

garethr:
@cmurphy thanks, rebased against current master

cmurphy:
@garethr what's the plan for the other tagged resources? e.g. https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_securitygroup.rb#L41 and https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/elb_loadbalancer.rb#L36 are params, not properties, but will they be changed to properties later and benefit from this patch?

Just want to clarify whether that is future work or if it belongs in this PR.

garethr:
@cmurphy tagging isn't fully implemented for other resources yet in the providers, and I have incoming in the VPC PR (https://github.com/garethr/puppetlabs-aws/commit/7d6ceca7417e73b980af789912285f39d7b6c59e). I'll consolidate tagging support later.

PR #54, opened 2015-01-19
This moves a few calls around to allow us to enable the pending tests

garethr:
This results in an additional call to the API, but we we're already
waiting for the running state in the after step so it's not much slower.


garethr:
This has been superseded by the wait now in create for instances.

PR #56, opened 2015-01-20
Prevent clash between blacksmith and metadata-lint due to gem dependency

tphoney:
unit tests failing due to dependencies in puppetblacksmith


garethr:
This relates to the issue described here https://github.com/maestrodev/puppet-blacksmith/issues/14

PR #57, opened 2015-01-20
Adding support for RDS instances. 

daveseff:
My first puppet provider module. It was a learning experience. I can now create and destroy RDS instances using puppet. Please code review. I'm sure I missed/buggered up something. :)

### DIFF
--- a/lib/puppet/type/rds_instance.rb
+++ b/lib/puppet/type/rds_instance.rb
@@ -31,19 +31,19 @@ Puppet::Type.newtype(:rds_instance) do
     end
   end
 
-  newproperty(:instance_id) do
-    desc 'the instance id to use for the instance'
+  newproperty(:db_instance_class) do
+    desc 'the type to use for the instance (mysql, | postgres)'
     validate do |value|
-      fail 'instance_id should not contain spaces' if value =~ /\s/
-      fail 'instance_id should not be blank' if value == ''
+      fail 'db_instance_class should not contain spaces' if value =~ /\s/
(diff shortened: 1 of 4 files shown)


lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,149 @@
...
+  ensurable do
+    defaultvalues
+    aliasvalue(:available, :present)
+    newvalue(:destroy) do
+      provider.destroy
+    end
+  end
+
+  newparam(:db_name, namevar: true) do
+    desc 'the name of the db instance'
```
petems: So this is deceptive: the db_name is not exactly the namevar: the db_name is the name of the database inside the instance:

```
DBName -> (string)
The meaning of this parameter differs according to the database engine you use. For example, this value returns either MySQL or PostgreSQL information when returning values from create-db-instance-read-replica since Read Replicas are only supported for MySQL and PostgreSQL.
MySQL, SQL Server, PostgreSQL
Contains the name of the initial database of this instance that was provided at create time, if one was specified when the DB instance was created. This same name is returned for the life of the DB instance.
Type: kms-key-id
Oracle
Contains the Oracle System ID (SID) of the created DB instance. Not shown when the returned parameters do not apply to an Oracle DB instance.
```

The namevar is actually `db_instance_identifier`

garethr:
Thanks @daveseff, just tidying up. Looks like this is continuing in https://github.com/puppetlabs/puppetlabs-aws/pull/69

PR #59, opened 2015-01-21
Allow the user to specify a IAM role during config

pjfoley:


Iristyle:
Thanks for the submission @pjfoley - any chance you could add corresponding tests to spec / acceptance at:

https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/unit/type/ec2_instance_spec.rb
https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/acceptance/instance_spec.rb

Thanks!

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -24,7 +24,7 @@ def self.instances
   end
 
   read_only(:instance_id, :instance_type, :region, :user_data, :key_name,
-            :availability_zones, :security_groups, :monitoring)
+            :availability_zones, :security_groups, :monitoring, :iam_role)
```
Iristyle: I think this should be `:iam_instance_profile` rather than `:iam_role`

pjfoley: Good pickup, I will make the change and add spec tests later today.

pjfoley:
Sorry it has taken so long to update.  I have fixed the incorrect parameter name noted above.

@Iristyle I dont have a lot of experience writing spec and acceptance tests.  I have just followed how the other spec tests are setup.  Let me know if I need to do more.

I am not sure what I need to do for the acceptance tests, do I have to somehow programmically create an IAM role and than have it used as part of the acceptance test to confirm that it works?  Any guidance would be appreciated.

Thanks,

Peter

lib/puppet/type/ec2_instance.rb
```diff
@@ -20,6 +20,10 @@
     desc 'the security groups to associate the instance'
   end
 
+  newparam(:iam_instance_profile) do
```
garethr: It would be nice if this where a property, as properties are displayed automatically by puppet resource.

garethr:
Hi @pjfoley, thanks for this. Looking closer I think the user interface might be better as two properties, rather than the hash. So something like:

```puppet
instance { 'something':
   iam_instance_profile_name => 'sample',
}
```

And:

```puppet
instance { 'something':
   iam_instance_profile_arn => 'arn...',
}
```

The reason for this is that the property names then map to a single thing, which means you don't need to find documentation on the structure of the hash. The current implementation also doesn't validate the hash structure, if you use two properties this problem goes away. Also, given a IAM profile has both a name and an ARN this would allow us to show both of those when we query for properties using puppet resource. Let me know what you think.

Regarding the acceptance tests, if it's easier I can stub out the tests and send you a PR? It might be easier if you rebase against master first as this no-longer merges cleanly. The tests for this would require an existing IAM profile to be created manually in the AWS account used to run the tests - at least until we can do that directly in Puppet.

Thanks again, and if you want to chat more directly just let me know.

pjfoley:
Ok, will work on this updating in line with the comments.  @garethr I will take a stab at the acceptance tests and if I run into issues seek your help.

I would also be interested in your view on a WIP pull request I just submitted to the puppet code base (specifically https://github.com/puppetlabs/puppet/pull/3675 ) I see these two items being related and beneficial for users of this module.

I am just starting to play around with setting up Puppet infrastructure on AWS.  I would like to keep my hiera-gpg key in a secured S3 bucket and use AWS IAM roles to grant access to pulling information out.  I have sketched out working code to extend the Puppet File resource type to accept S3 urls (s3://bucket-name/file-name).

Happy to take your advice if this code could be something that could be embedded in the this module.

pjfoley:
Hi @garethr I have added properties to my local code base, I am trying to work out how I can programmically restrict enforce the user using either the IAM role name property or the IAM arn property only. I don't want a user to set them both at the same time (ie how do I check if the name an arn match the same IAM role?).

Could you point my to an example either in this module or in the puppet codebase of an example on this restriction?

Thanks.

garethr:
Hi @pjfoley. Your best bet I think is a type level validator, here's a similar example: https://github.com/puppetlabs/puppetlabs-aws/pull/93/files#diff-5696adcac12ae25a8fdf462f918f2f8cR80

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -70,6 +71,7 @@ def self.instance_to_hash(region, instance)
       tags: tags,
       region: region,
       hypervisor: instance.hypervisor,
+      iam_instance_profile_arn: instance.iam_instance_profile ? instance.iam_instance_profile.arn : nil,
```
pjfoley: The instance response returned from the amazon api does not provide the IAM role friendly name.

I have looked at the Aws::IAM::Client to see if I can get the friendly name but have been unable to find a method.  The closest I could work out (which may be over kill) would be to get a list of all the roles the user can query against and than look up the arn returned to find the friendly name.

Something like:
```ruby
m_iam = Aws::IAM::Client.new(region:'ap-southeast-2')
iam_instance_profile_arn = m_iam.list_instance_profiles.instance_profiles.select {|item| item.arn == 'arn:aws:iam::#########:instance-profile/???????????'}.map(&:instance_profile_name)
```
Which is not ideal.

Thoughts?

pjfoley:
@garethr I have looked through the acceptance tests already there, I see what I would need to do.

I am happy to add some acceptance tests can you create an IAM role under the AWS account travis builds against.  Can you let me know the ARN string and the friendly name.

Also any pointers on how I can ensure that the tests only run via Travis?  or at least provide context for the failure for any developers who run the tests from their own AWS accounts?

garethr:
@pjfoley I think the best pattern for the moment is using an ENV variable to pass in the details, we do this already here: https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/acceptance/instance_spec.rb#L376

The acceptance tests don't run in Travis, but on our internal CI system so as to better control the credentials and spend. They also only run when you run `rake acceptance` so it's fine to be part of the main set. If the ENV variable isn't passed, for the moment we can just skip the test.

lib/puppet/provider/ec2_instance/v2.rb
garethr: Mmm, so it might make sense here for the `iam_instance_profile_arn` to be a property (as you have it) and the `iam_instance_profile_name` to be a param (ie. it's only passed in on create). That would mean we don't need it back from the API. I think at least to begin with that's good enough.

spec/acceptance/instance_spec.rb
```diff
@@ -35,6 +35,13 @@ def get_instance(name)
         :volume_size => 8,
       }
 
+
+      # The value for this ENV var must be an existing IAM role in your Amazon account
+      # the IAM role must be available in the acceptance test region.
+      #
+      # When testing IAM roles you need to provide the IAM role name and the coresponding ARN.
+      @config[:iam_instance_profile_name] = ENV['IAM_ROLE_NAME'] if ENV['IAM_ROLE_NAME'] && ENV['IAM_ROLE_ARN']
```
pjfoley: @garethr I am interested in your view on if you think it is necessary to test both the iam_instance_profile_name and iam_instance_profile_arn code paths when creating an instance?

Currently I have the acceptance test setup to create the instance with the value from iam_instance_profile_name and validate the IAM role based on the value of the iam_instance_profile_arn.

This would save provisioning an additional instance just to confirm setting the IAM role via the iam_instance_profile_arn works.

Thoughts?

pjfoley:
@garethr happy to hear your view on the acceptance test and if it meets your standard to merge into master yet.

spec/acceptance/instance_spec.rb
garethr: I think I'd probably prefer to test both. It's likely we'll do some consolidation of the acceptance tests at some point in the future, but I'd prefer to err on the side of better coverage for the moment.

pjfoley:
@garethr just checking if you needed anything else from me?

garethr:
@pjfoley sorry for the delay, I've been a touch busy unfortunately. I'll try and have a proper look at this in the next week or so. From a brief pass it's looking good. I might change some of the README additions as we're currently producing more comprehensive documentation, and I'd probably squash some of the commits down, but if it's just that I'll do that before merging.

Thanks again for taking a run at this. Much appreciated.

pjfoley:
Thats ok, I understand it is busy with the 4.0 crunch you guys are under. I just wanted to make sure you were not waiting for me to do something else.  Sing out if there is anything I need to do.

xaptronic:
Is there anything blocking this from being merged?

garethr:
@xaptronic just finding time to get to a number of things, apologies for the delay. I'm on holiday at the moment but will definitely get to in the next few weeks.

xaptronic:
By all means sir! Enjoy the holidays!

garethr:
@pjfoley thanks again for your work on this. I've just made a few minor changes and merged in now.

pjfoley:
@garethr no problem happy to help.

PR #63, opened 2015-01-26
Remove additional calls to API for restart, stop and destroy

garethr:
In the majority of cases this will have the same result, just with fewer
calls to the API - by using the already known instance_id.

It also prevents the odd behaviour of stopping/restarting multiple
instances in the case where you had multiple instances with the same
name. It moves all the error handing requirements to the prefetch.


cmurphy:
+1 but needs a rebase.

garethr:
@cmurphy thanks, rebased against current master

PR #68, opened 2015-02-05
mutable property tests for autoscaling group

zreichert:
Work in process do not merge
This PR lacks the ability to teardown its own security groups, functionality contained in https://github.com/puppetlabs/puppetlabs-aws/pull/66
Test that are commented out are duplicates of tests in the same file, need to determine which location they should be in.


spec/acceptance/autoscaling_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'spec_helper_acceptance'
 require 'securerandom'
+require 'pry'
```
zreichert: oops will remove

garethr:
Note that I just merged #66

garethr:
@zreichert I notice this still says work in progress in the comment? It also has commented out tests, if they're future functionality then maybe mark as pending?

zreichert:
@garethr I'll make those functional.  There is also an un implemented test.  Those commented tests are duplicates of existing tests that I realized after the fact.  Need to decide which one to keep.

spec/acceptance/autoscaling_spec.rb
```diff
@@ -32,39 +32,64 @@ def find_alarm(name)
...
+      duplicate_delete = 'duplicate_asg_delete.pp.tmpl'
+      r = PuppetManifest.new(@asg_template_delete, @asg_config).apply
+      # assert that none of the results contain 'Error:'
+      expect(r[:output].any?{ |o| o.include?('Error:')}).to eq(false)
       response = @aws.autoscaling_client.describe_auto_scaling_groups(
-        auto_scaling_group_names: ["#{@name}-asg"],
+        auto_scaling_group_names: [@asg_config[:asg_name]],
       )
       id = Array.new
       response.to_h[:auto_scaling_groups].first[:instances].each do |x|
```
garethr: While running this I noticed that this line fails with an exception, but doesn't cause the tests to fail as it's in the after block. It was added in https://github.com/puppetlabs/puppetlabs-aws/commit/e7ae82830ae108a15eb5afc5dbe8e235e065f334

Specifically, to_h fails with `TypeError: wrong element type Aws::PageableResponse (expected array)`. The fix is to access the data directly like so:

```ruby
response.data[:auto_scaling_groups].first[:instances].each do |x|
```

garethr:
@zreichert this looks good to me and runs cleanly, apart from the issue in the comment, if you could fix that I'll merge.

zreichert:
@garethr Thanks!  Should be good to go now.

PR #69, opened 2015-02-06
(CLOUD-295) Add RDS Support

petems:
Rebase of https://github.com/puppetlabs/puppetlabs-aws/pull/57 with spec tests added :+1: 

@daveseff want to have a look? :+1:


lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -0,0 +1,102 @@
+require_relative '../../../puppet_x/puppetlabs/aws.rb'
+require "base64"
```
garethr: Not sure this is used in this provider?

lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -0,0 +1,102 @@
...
+      if resource = resources[prov.name] # rubocop:disable Lint/AssignmentInCondition
+        resource.provider = prov if resource[:region] == prov.region
+      end
+    end
+  end
+
+  def self.db_instance_to_hash(region, instance)
+    tags = rds_client(region).list_tags_for_resource(resource_name: get_arn_for_instance(region,instance.db_instance_identifier))
+
+    config = {
```
garethr: You can remove the variable assignment here to config and just return the hash.

lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -0,0 +1,102 @@
...
+      storage_type: resource[:storage_type],
+      multi_az: resource[:multi_az],
+      allocated_storage: resource[:allocated_storage],
+      iops: resource[:iops],
+      master_username: resource[:master_username],
+      master_user_password: resource[:master_user_password],
+      db_subnet_group_name: resource[:db_subnet_group_name],
+    }
+
+    response = rds_client(resource[:region]).create_db_instance(config)
```
garethr: You can remove the assignment to response here.

lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    @property_hash[:ensure] = :present
+  end
+
+  def destroy
+    Puppet.info("Deleting instance #{name} in region #{resource[:region]}")
```
garethr: should be deleting database or similar here. Also see other logging messages.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    defaultvalues
+    aliasvalue(:available, :present)
+    newvalue(:destroy) do
+      provider.destroy
+    end
+  end
+
+  newparam(:name, namevar: true) do
+    desc 'the name of the db instance (also known as the db_instance_identifier)'
+    validate do |value|
```
garethr: Can remove this blank block

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
+Puppet::Type.newtype(:rds_instance) do
+  @doc = 'type representing an RDS instance'
+
+  ensurable do
+    defaultvalues
+    aliasvalue(:available, :present)
+    newvalue(:destroy) do
```
garethr: What's the rationale for introducing destroy directly and not just using absent?

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
+Puppet::Type.newtype(:rds_instance) do
+  @doc = 'type representing an RDS instance'
```
garethr: We're moving all of the descriptions to leading caps and a full stop at the end to match the default description for ensure. If you could switch them over. Very picky.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+
+  newproperty(:db_instance_class) do
+    desc 'the type to use for the instance (mysql, | postgres)'
+    validate do |value|
+      fail 'db_instance_class should not contain spaces' if value =~ /\s/
+      fail 'db_instance_class should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:availability_zone_name) do
```
garethr: I think based on other types this should just be `availability_zone` https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb#L76

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+  newproperty(:storage_type) do
+    desc 'The storage stype for the DB (Valid values: gp | io1  *Note: If you specify io1, you must also include a value for the Iops parameter)'
+    validate do |value|
+      fail 'storage_type type should not contains spaces' if value =~ /\s/
+    end
+  end
+
+  newproperty(:iops) do
+    desc 'The IOPS stype for the DB (minimum 1000)'
+    validate do |value|
```
garethr: Does this need to be an integer? We could validate for that here.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+
+  newproperty(:multi_az) do
+    desc 'The main user Password'
+    validate do |value|
+      fail 'multi_az should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:security_groups) do
+    desc 'The main user Password'
```
garethr: Wrong description

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+  end
+
+  newproperty(:multi_az) do
+    desc 'The main user Password'
+    validate do |value|
+      fail 'multi_az should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:security_groups) do
```
garethr: This probably wants to work similar to instances as an array: https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb#L36

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    desc 'The main user Password'
+    validate do |value|
+      fail 'security_groups should not be blank' if value == ''
```
garethr: Is this true? In many cases the default security group is used if one is not provided? Is that not true for RDS?

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    end
+  end
+
+  newproperty(:db_subnet_group_name) do
```
garethr: I think this just wants to be subnet, to match https://github.com/garethr/puppetlabs-aws/blob/add-subnet-support-to-instance/lib/puppet/type/ec2_instance.rb#L124. Once the VPC support is merged this can also have an autorequire on the subnet type.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    desc 'The VPC subnet for this instance.'
+    validate do |value|
+      fail 'db_subnet_group_name should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:skip_final_snapshot) do
```
garethr: The name of the property is negative (skip) and the description positive. Also, is this a boolean? In which case it should probably be a boolean type. See https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/ec2_instance.rb#L52 for an example.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    desc 'Create one last snapshot.'
+    validate do |value|
+      fail 'skip_final_snapshot should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:final_db_snapshot_identifier) do
+    desc 'Create one last snapshot.'
```
garethr: Description doesn't match property name

tests/create.pp
```diff
@@ -83,3 +83,19 @@
     port     => 80,
   }],
 }
+
+rds_instance { 'db-name-5':
```
garethr: Instead of placing these in the simple test file could you create an example in the examples directory? See the other examples for an, er, example.

README.md
```diff
@@ -224,6 +224,27 @@ elb_loadbalancer { 'name-of-load-balancer':
 }
 ```
 
+### rds_instance
```
garethr: Just include this in the examples directory rather than here for the moment.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+
+Cannot be longer than 8 characters
+SQL Server
+
+Not applicable. Must be null.'
+    validate do |value|
+    end
+  end
+
+  newparam(:security_groups, :array_matching => :all) do
```
garethr: This should ideally be a property, assuming the name is available in the describe_db_instances call. (I know for instances it's not at the moment, I have that fixed locally)

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,184 @@
...
+    end
+  end
+
+  newparam(:name, namevar: true) do
+    desc 'the name of the db instance (also known as the db_instance_identifier)'
+    validate do |value|
+    end
+  end
+
+  newparam(:db_name) do
```
garethr: Any reason for this being a param rather than a property? I see you populate the property_hash with the value and it feels useful to puppet resource?

garethr:
Thanks @petems @daveseff. I've left lots of comments, mainly small things. We'll need to expand the unit tests and add some acceptance tests to this but it's looking good. I'm happy to help out if I can.

petems:
@garethr Made the changes required based on comments, I can rebase the commits later on :+1:

lib/puppet/type/rds_instance.rb
petems: Fixed

lib/puppet/type/rds_instance.rb
petems: Changed to `defaultto :false`

tests/create.pp
petems: Something like:

Create RDS MySQL instance, connect to instance using puppetlabs-mysql?

garethr: Even just creating the instance would be fine. It would be great to have the connection example though as well.

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,181 @@
...
+
+  newproperty(:storage_type) do
+    desc 'The storage stype for the DB (Valid values: gp | io1  *Note: If you specify io1, you must also include a value for the Iops parameter)'
+    validate do |value|
+      fail 'storage_type type should not contains spaces' if value =~ /\s/
+    end
+  end
+
+  newproperty(:iops) do
+    desc 'The IOPS stype for the DB (minimum 1000)'
```
petems: From reading the docs, by default this is 0, and that will set your RDS as standard storage rather than the Provisioned IOPS. So I'll probably remove the comment about minimum

daveseff:
Any news on when this will get merged in?

petems:
@daveseff Me and Gareth are going to do some acceptance test pairing soonish, so hopefully should be able to get it tied up soon! :+1:

daveseff:
Nice. Can I remove my fork of this then?

lib/puppet/type/rds_instance.rb
petems: Changed back to default to :true

petems:
@daveseff Yep, you can probably delete your fork, your commits are safe and sound in this branch! :+1:

danieldreier:
Thanks for your work adding RDS support.

The way I'd like to use the RDS support is to create a multi-AZ MySQL database in the production environment, with backups, maintenance window set, and a parameter group defined, in our VPC.

The only major thing that I noticed about that RDS PR is the inability to manage parameter groups, which are the only mechanism for Amazon gives for configuring MySQL settings. Drupal is so bad about making huge JOINs that tuning mysql really isn't optional - the default settings will have it constantly writing temp tables to disk and performance is significantly impacted. 

The most basic use case would be simply associating an RDS instance with a parameter group, though it would really be nice to create parameter groups as well. It would be really ideal if I could copy-paste the settings from the puppetlabs-mysql module into a parameter group. It's just a hash of settings, so it seems like it should be possible.

petems:
@danieldreier Just added Parameter Group type and provider, and added being able to add it to an RDS instance on creation :+1:

danieldreier:
awesome! That was quick. Thanks!

choonming:
I was looking into the code and the API documentation on RDS. Noted that you only allow creating DB security groups with some manual work involved as per your documentation in your branch. Is it possible to add a parameter to define the VPC security group associated with the new RDS instance?

petems:
@choonming Right now the AWS API is limited to just creating and deleting DB security groups. You have to do the assignment of EC2 or VPC security groups through the GUI.

petems:
@choonming Actually, I'm not positive on that for VPC groups, the docs page is timing out for me right now, I'll double check that later! :+1:

examples/postgres-rds-example/README.md
```diff
@@ -0,0 +1,53 @@
+# RDS
+
+[Amazon Relational Database Service](http://aws.amazon.com/rds/) (Amazon RDS) is a web service that makes it easy to set up, operate, and scale a relational database in the cloud.
+
+## How
+
+This example creates a security group to allow Postgres RDS instance, then creates an RDS instance with that security group assigned.
+
+  puppet apply rds_security.pp
```
garethr: Needs a few extra leading spaces.

lib/puppet/provider/rds_db_securitygroup/v2.rb
```diff
@@ -0,0 +1,93 @@
...
+    }
+  end
+
+  def exists?
+    Puppet.info("Checking if DB Security Group #{name} exists")
+    [:present, :creating, :available].include? @property_hash[:ensure]
+  end
+
+  def create
+    Puppet.info("Starting DB instance #{name}")
```
garethr: This is creating a security group not a DB instance I think?

lib/puppet/type/rds_instance.rb
```diff
@@ -0,0 +1,172 @@
...
+  end
+
+  newproperty(:multi_az) do
+    desc 'Define a multi-az'
+    validate do |value|
+      fail 'multi_az should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:db_subnet_group_name) do
```
garethr: I think this should just be called subnet to match other resources that do the same. Everything is referenced by name so `group_name` is redundant. and `db` is redundant as it's part of the database type.

garethr:
@peter This is looking good, apologies for taking so long to get to it. I've left a few more minor comments. On top of those we'll need:

* Details in the README for the types and properties. See the examples from the 1.0 README
* Unit tests covering the type property inputs (rather than just existence). This probably involves adding basic type checking as a minimum to the properties - again see lots of examples from the existing types.

choonming:
@petems if we look at the RDS API documentation on ModifyDBInstance, we can specify the following:

VpcSecurityGroupIds.member.N
A list of EC2 VPC security groups to authorize on this DB instance. This change is asynchronously applied as soon as possible.
    Constraints:
        Must be 1 to 255 alphanumeric characters
        First character must be a letter
        Cannot end with a hyphen or contain two consecutive hyphens
    Type: String list
    Required: No

petems:
@choonming Ok, I'll see if I can add that to the creation parameters :+1:

lib/puppet/type/rds_instance.rb
petems: Fixed :+1:

lib/puppet/provider/rds_db_securitygroup/v2.rb
petems: Fixed :+1:

examples/postgres-rds-example/README.md
petems: Fixed :+1:

Iristyle:
PR references `CLOUD-295` but commits all reference `MODULES-1702`

petems:
Yep, need to rebase and rename the commits to the new ticket name.

petems:
Closed in favour of new ticket name branch: https://github.com/puppetlabs/puppetlabs-aws/pull/161

PR #71, opened 2015-02-09
(CLOUD-196) Expose instance security groups to puppet resource

garethr:


spec/unit/type/ec2_instance_spec.rb
```diff
@@ -7,13 +7,14 @@
...
       :name,
-      :security_groups
     ]
   end
 
   let :properties do
     [
       :ensure,
+      :security_groups,
+      :tags,
```
cmurphy: Is this relevant to this PR?

garethr: Note the tags line. I can push that into master directly if you'd prefer.

Iristyle: We scrapped `tags` from this commit... so feel free to put it in #53 or add a `(maint)` commit to master if you'd like.

PR #74, opened 2015-02-11
(CLOUD-204, CLOUD-161) Work around failures to tag new instances

garethr:
Due to either network failures (over the internet) and/or eventual
consistency issues in AWS, it's possible to start an instance and then
fail to tag it. The API doesn't expose a tranactional endpoint for this.

Unfortunately, names of instances are just tags. So in the case tagging
fails we can't retrieve information via name later.

We have tried simply waiting for the instance to be running (which
minimises the eventual consistency issues) but have still seen the odd
failure. This patch replaces that wait (a good thing, instance creation
is now quicker from Puppet's point of view) with retries of the tagging
calls in case of failure.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -11,6 +11,7 @@ def location_for(place, fake_version = nil)
 end
 
 gem 'aws-sdk-core', '2.0.5'
+gem 'retries'
 
 group :test do
   gem 'rake'
--- a/README.md
+++ b/README.md
@@ -32,16 +32,20 @@ Puppet.
 
     gem install aws-sdk-core
 
-If you're running Puppet Enterprise you need to install the gem using
+We also use the retries library so install that with:
+
+    gem install retries
+
+If you're running Puppet Enterprise you need to install the gems using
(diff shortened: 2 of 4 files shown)


cmurphy:
Is it necessary to add a new gem dependency for this? A rescue/retry loop would only be a few extra lines.

Does create_tags emit a specific exception the retry can rescue? Is there ever a case where it might raise a different exception that indicates a retry would be unnecessary or unhelpful?

garethr:
@cmurphy The gem provides exponential back off, other features that will be useful and a nice abstraction. It's not a large amount of code, so we could just vendor it, but I don't see the downside of requiring the gem (especially as we already have a gem dependency) and the upside avoids the maintenance overhead of vendoring (at least at this stage).

I thought about being more explicit with the exception handling too, but their are a range of problems that could occur that retrying fixes, and I couldn't see a case where it would be unhelpful to retry.

cmurphy:
Sounds good to me.

PR #77, opened 2015-02-17
(MAINT) Handle username of PE masters > 3.7.0

petems:


garethr:
Thanks. I'd seen the email be truncated, so the README was correct.

PR #78, opened 2015-02-19
(CLOUD-203) Support tagging for security groups

garethr:
This adds the ability to view tags using puppet resource and to
chnage tags after the security group has been created.

This should be rebased and merged after the VPC PR as it relies on the
tagging helpers containers in that work. Note this means a few specs will fail
until that happens.

### DIFF
--- a/spec/unit/type/ec2_securitygroup_spec.rb
+++ b/spec/unit/type/ec2_securitygroup_spec.rb
@@ -7,7 +7,6 @@ describe type_class do
   let :params do
     [
       :name,
-      :tags,
     ]
   end
 
@@ -17,6 +16,7 @@ describe type_class do
       :description,
       :region,
       :ingress,
+      :tags,
     ]
   end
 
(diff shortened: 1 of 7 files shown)


Iristyle:
Spec fails require updating.

spec/unit/type/ec2_securitygroup_spec.rb
```diff
@@ -31,4 +31,14 @@
       expect(type_class.parameters).to be_include(param)
     end
   end
+
+  it 'should order tags on output' do
```
Iristyle: I wonder if this test should be moved to a shared location give its the same basic test for all resources supporting tags?

garethr: Me and @zreichert had been talking about moving a few things into customer rspec matchers and the like. Agree this is a prime candidate.

garethr:
@Iristyle rebased to bring in the mentioned helpers from the VPC work, specs should be passing now.

I've also moved the duplicate test code into a matcher as discussed, and used that for the resources that previously had that test in place.

PR #84, opened 2015-02-23
(CLOUD-231) Additional instance properties, including support for block devices

garethr:


spec/acceptance/fixtures/instance.pp.tmpl
```diff
@@ -9,6 +9,10 @@ ec2_instance { '{{name}}':
     {{k}} => '{{v}}',
   {{/tags}}
   },
+  block_devices => [{
+    device_name => '{{device_name}}',
+    volume_size => {{volume_size}},
```
cmurphy: Tests are failing here when volume_size is not set.

garethr: Ah, it was also used in the loadbalancer acceptance test and I'd missed one of the config hashes. Should be fixed now.

cmurphy:
Rebased and merged to master.

PR #86, opened 2015-02-24
Fixes Missing bracket

petems:
We should probably put the examples into CI too, at least a lint or syntax check?


garethr:
@petems agreed, it's been in a list in my head for a while. Just not at the top. Thanks for the fix.

PR #87, opened 2015-02-24
(CLOUD-232) Add VPC security group support

hunner:
The security groups in VPCs must be managed by their IDs, so I added `id` to the ec2_securitygroup type and updated the provider to only deal with the IDs.

The `vpc_name` property for ec2_securitygroup was added to create a SG in a VPC. This may only be done on SG creation, so I created a setter method that raises an error if it is not insync after create.

The ec2_vpc type now returns its ID. The provider was already setting it in `instances`.

This PR also contains a required bugfix.

### DIFF
--- a/lib/puppet/type/ec2_securitygroup.rb
+++ b/lib/puppet/type/ec2_securitygroup.rb
@@ -49,6 +49,17 @@ Puppet::Type.newtype(:ec2_securitygroup) do
     end
   end
 
+  newproperty(:vpc_name) do
+    desc 'The name of the virtual private cloud (VPC) the security group belongs to, if applicable.'
+  end
+
+  newproperty(:id) do
+    desc 'The AWS ID of the security group'
+    validate do |value|
+      fail "id is read-only"
+    end
+  end
+
   def should_autorequire?(rule)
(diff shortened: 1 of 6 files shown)


hunner:
Needs docs and tests.

lib/puppet/type/ec2_securitygroup.rb
```diff
@@ -49,6 +49,17 @@ def stringify_values(rules)
     end
   end
 
+  newproperty(:vpc_name) do
+    desc 'The name of the virtual private cloud (VPC) the security group belongs to, if applicable.'
+  end
+
+  newproperty(:id) do
```
Iristyle: I would probably separate out the addition of the `:id` property for `ec2_securitygroup` to it's own commit as that's a standalone change.  Then adding `vpc_name` can be pared back to a smaller set of changes.

garethr:
I'm going to close this in favour of https://github.com/puppetlabs/puppetlabs-aws/pull/88/. I've ported the fix in https://github.com/hunner/puppetlabs-aws/commit/62bfa1224688f5f7025d177da7142a938fc2a90b across as well, thanks for that.

PR #88, opened 2015-02-24
(CLOUD-232) Support for adding security groups to VPC

garethr:
This maintains backwards compatibillity with the existing security group
interface, but works with both VPC and non-vpc security groups.

### DIFF
--- a/examples/vpc-example/init.pp
+++ b/examples/vpc-example/init.pp
@@ -4,6 +4,16 @@ ec2_vpc { 'sample-vpc':
   cidr_block   => '10.0.0.0/16',
 }
 
+ec2_securitygroup { 'sample-sg':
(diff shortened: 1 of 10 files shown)


daveseff:
I Like this. I actually implemented this same thing on my local copy. I think you beat me to it. 

BTW what is the function of all the vcr_cassettes? I've seen the files but do not understand what they do.

garethr:
Hi @daveseff. Thanks. The cassettes are fixtures for the provider tests, using the VCR (https://github.com/vcr/vcr) library. I'd like to look at more explicit mocking at some point to avoid the need for them, but in the short term it's a nice easy way of isolating those tests from the network. We have the full acceptance tests suite for that.

PR #89, opened 2015-02-25
(CLOUD-233) Add the ability to associate elastic IP addresses with instances

garethr:
This doesn't provide the ability to allocate IP addresses, just
associate them.

Allocation is difficult within Puppet, as you can't know the IP address
ahead of time, and the Elastic IP resource doesn't support any form of
metadata or tags.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -21,3 +21,5 @@ infrastructure.
   Puppet DSL to manage a AWS VPC environment
 * [Using IAM permissions](iam-profile) - an example IAM profile for
   controlling the API permissions required by the module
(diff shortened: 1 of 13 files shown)


hunner:
LGTM, but should get some acceptance tests before merging probably. (Using the cli to create the EIPs for the tests and teardown at the end.)

garethr:
@hunner thanks, I'll add some tests if were happy with the approach, probably using the API to allocate and throw away EIPs from the test code.

garethr:
@hunner just a note that I've added some acceptance tests to this to demonstrate IP allocation.

PR #91, opened 2015-02-26
(CLOUD-122) Adds support for launching auto scaling groups in VPC

garethr:


examples/auto-scaling-groups/README.md
```diff
@@ -41,8 +41,32 @@ security group with:
...
+ec2_autoscalinggroup { 'test-asg':
+  ...
+  subnets => 'subnet-name',
+}
+```
+
+Note that the security groups attached to the launch configuration must
+be associated with the same subnet. In cases where multiple security
+groups exist with the same name, typically the case for the default
+ssecurity group, you can provide a hint to the launch configuration like
```
cmurphy: Typo - ssecurity

lib/puppet/type/ec2_launchconfiguration.rb
```diff
@@ -52,9 +52,17 @@ def insync?(is)
     end
   end
 
+  newparam(:vpc) do
+    desc 'A hint for detecting ambiguously named groups liked default.'
```
cmurphy: Seems like a typo (liked/like) but even corrected this could be a little more verbose.

cmurphy:
The acceptance tests fail consistently for me on a non-vpc-only account. Trying it manually, it seems like the launchconfiguration provider is having trouble finding the security group.

cmurphy:
Actually, that might be related to an existing CI issue in master. However this still produces an error:

```
Error: Could not prefetch ec2_launchconfiguration provider 'v2': Invalid id: "test-sg" (expecting "sg-...")
```

which might be causing additional failures.

examples/auto-scaling-groups/README.md
garethr: Thanks, fixed.

lib/puppet/type/ec2_launchconfiguration.rb
garethr: Thanks, fixed.

garethr:
@cmurphy interesting. So the issue was that an existing launch configuration had a security group of 'test-sg' which appeared to have been deleted manually. This meant the API returns the name, not the id, which caused an exception Aws::EC2::Errors::InvalidGroupIdMalformed. I'm not catching that edge case.

cmurphy:
@garethr I'm still seeing the following error during the autoscaling tests:

```
Error: Could not prefetch ec2_launchconfiguration provider 'v2': The security group 'sg-4517ab5a' does not exist
```

garethr:
Humf. This this was a similar action at a distance issue, where the security group identifier wasn't invalid but was similarly detached - leading to a different and uncaught exception. We're now catching GroupNotFound as well as Invalid.

PR #75, opened 2015-02-12
(CLOUD-208, CLOUD-209, CLOUD-210, CLOUD-47) Improve puppet resource support for ELB load balancer

garethr:
* Show instances attached to ELBs in puppet resource
* Show availability zones for ELBs in puppet resource
* Show ELB listeners in puppet resource

This enables a number of pending tests of the above functionality, as
well as adds tests in for the ELB type properties.

Also, security groups for ELBs are only present for VPC, which we don't
explicitly support yet. For the moment I've removed the property.

PR #93, opened 2015-03-02
(CLOUD-211) Allow load balancers to be created in VPC environments

garethr:
This allows for:

* attaching a ELB to a list of subnets
* associating an ELB with security groups (a VPC only feature)
* specifying whether the ELB is internal only or has a public address

This also updates the descriptions on the ELB type to match the format elsewhere, Leading caps and a full stop at the end.

### DIFF
--- a/lib/puppet/provider/elb_loadbalancer/v2.rb
+++ b/lib/puppet/provider/elb_loadbalancer/v2.rb
@@ -18,7 +18,7 @@ Puppet::Type.type(:elb_loadbalancer).provide(:v2, :parent => PuppetX::Puppetlabs
     end.flatten
   end
 
-  read_only(:region, :scheme, :availability_zones, :listeners)
+  read_only(:region, :scheme, :availability_zones, :listeners, :tags)
 
   def self.prefetch(resources)
     instances.each do |prov|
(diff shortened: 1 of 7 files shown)


cmurphy:
This has test failures for me:

```
Error: SecurityGroups must not be empty
```

Seems like the elb_loadbalancer needs to have security_groups defined as a parameter in the vpc acceptance test.

garethr:
@cmurphy thanks, should be resolved. The cause was not dealing correctly with setting a default blank array, and an the AWS side that defaulting to ['default'] - and hence triggering an out of sync error.

cmurphy:
This needs a rebase since #75 was merged first.

garethr:
@cmurphy just a note to say I rebased this against master. It should cleanly merge now.

cmurphy:
This is now failing acceptance again, maybe an error in the rebase caused an issue in the template?

```
Wrapped exception:
listeners must include load_balancer_port
```

garethr:
@cmurphy thanks. We'd added an ELB to the VPC tests but change the interface. Now fixed. This handily also showed up an issue with syncing tags for ELBs which is also now fixed.

daveseff:
Maybe I'm missing something.. The docs aren't clear. 

I've tried
`subnets  => [...], `
But got:
`Error: Failed to apply catalog: Validation of Elb_loadbalancer[foobar] failed: You can specify either subnets or availability_zones for the ELB foobar at` 

Then I tried, as per the doc:
`security_groups      => ['some-sg-name',],`
and got:

    Error: Could not run Puppet configuration client: Puppet detected a problem with the information returned from AWS
    when looking up elb_loadbalancer in us-east-1. The specific error was:
    
    The subnet ID 'subnet-7c616246' does not exist
    
    Rather than report on elb_loadbalancer resources in an inconsistent state we have exited.
    This could be because some other process is modifying AWS at the same time.

PR #95, opened 2015-03-04
Designate load balancers in autoscaling groups

brianwhigham:


lib/puppet/type/ec2_autoscalinggroup.rb
```diff
@@ -62,6 +62,13 @@ def insync?(is)
     end
   end
 
+  newproperty(:load_balancers_names, :array_matching => :all) do
```
garethr: I think this should be called `load_balancers` to match the naming convention elsewhere, Also, you'll want to set an `autorequire` to point to the `elb_loadbalancer` type, in order for Puppet to know that it should create/act on the load balancer before the autoscaling group if both are described in Puppet.

garethr:
@briancain thanks for this. It would be useful to add some basic tests for this too - in particular to the acceptance tests at https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/acceptance/autoscaling_spec.rb. If you'd like a hand with that let me know.

briancain:
hey @garethr I think you meant to tag @brianwhigham in this comment :)

mentat:
Hi there, I'm going to close this since it's been inactive for so long.  Please feel free to re-open in the future.

PR #97, opened 2015-03-05
(CLOUD-226 to CLOUD-228) Tests for ec2_vpc types

zreichert:
This PR covers
CLOUD-226
CLOUD-227
CLOUD-228

### DIFF
--- a/spec/acceptance/vpc_spec.rb
+++ b/spec/acceptance/vpc_spec.rb
@@ -206,40 +206,641 @@ describe "The AWS module" do
       expect(instance.vpc_id).to eq(@vpc.vpc_id)
     end
 
-    it 'should allow tags to be changed' do
-      expect(@aws.tag_difference(@vpc, @config[:tags])).to be_empty
-      tags = {
-        :department => 'engineering',
-        :created_by => 'aws-acceptance',
-        :foo => 'bar',
-      }
-      new_config = @config.dup.update(tags)
-      PuppetManifest.new(@template, new_config).apply
-      vpc = find_vpc("#{@name}-vpc")
-      expect(@aws.tag_difference(vpc, new_config[:tags])).to be_empty
+    context 'change tags' do
+      before(:all) do
+        expect(@aws.tag_difference(@vpc, @config[:tags])).to be_empty
+        expect(@aws.tag_difference(@option, @config[:tags])).to be_empty
+        expect(@aws.tag_difference(@subnet, @config[:tags])).to be_empty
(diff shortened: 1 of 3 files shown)


spec/acceptance/vpc_spec.rb
```diff
@@ -219,27 +219,544 @@ def generate_ip
...
+
+      it 'deletes the ec2_vpc_internet_gateway' do
         ENV['AWS_REGION'] = @default_region
-        options = {:name => "#{@name}-vpc"}
-        @result = TestExecutor.puppet_resource('ec2_vpc', options, '--modulepath ../')
+        name = "#{@delete_me[:name]}-igw"
+        options = {:name => name, :region => @default_region, :ensure => 'absent'}
+        result = TestExecutor.puppet_resource('ec2_vpc_internet_gateway', options, '--modulepath ../')
+        expect(result.stderr).not_to match(/Error/)
+        exspect(find_internet_gateway(name)).to raise_error
```
garethr: Typo, presumably should be `expect`

spec/acceptance/vpc_spec.rb
```diff
@@ -219,27 +219,544 @@ def generate_ip
...
+
+    end
+
+    context 'delete with puppet resource' do
+
       before(:all) do
+        @delete_me = @config.clone
+        @delete_me[:name] = "#{PuppetManifest.env_id}-#{SecureRandom.uuid}"
+        result = PuppetManifest.new(@template, @delete_me).apply
+        require 'pry'; binding.pry
```
garethr: Left in Pry call

spec/acceptance/vpc_spec.rb
```diff
@@ -219,27 +219,544 @@ def generate_ip
       expect(@aws.tag_difference(vpc, new_config[:tags])).to be_empty
     end
 
-    describe 'using puppet resource on the VPC' do
+  end
+
+  describe 'createing a new VPC environment with all possible properties' do
```
garethr: Typo: createing rather than creating

garethr:
@zreichert I've left a few small comments. Also, looking at the tests in `delete with puppet resource`, do these have an implied order in which they need to happen? For instance if I ran the delete route table test in isolation would it fail because the gateway it's attached to is still present? The other VPC tests work around this by using the relationships in the _delete manifest template.

zreichert:
@garethr Correct there is an implied order.  I could just merger these all in to one giant it block.  This would solve the ordering problem but the report would not be as clean.

zreichert:
@garethr Tests fixed.  Please review again.

PR #98, opened 2015-03-05
(maint) vpc_only_account? logic improvements

Iristyle:
- Clean up code to provide a more self-descriptive mechanism
   for finding VPC only accounts.

   With clasic accounts, account_types is ["EC2", "VPC"] while
   a VPC only account has the values ["VPC"].  It's possible
   that there could be other / legacy account types that have
   a different list of supported-platforms, or that this
   changes in the future.


Iristyle:
@zreichert / @garethr - quick PR on how vpc account detection is being performed.  Related, I think this should be memoized, since a Puppet run will only ever be run with a single AWS account.

zreichert:
+1
This is a good model for determining account capabilities.  AWS may add new account types in the future.

PR #99, opened 2015-03-06
(CLOUD-179) Avoid throwing exceptions in prefetch

garethr:
I have a guess this accounts for at least some orphaned resources left
after test runs. The story goes like this:

* Get a list of a resources (say instances)
* For each of those instances, we trigger further requests to get
  the names from the ids (say for related security groups)
* This was previously done using the *_ids param to describe_*
* This formulation will throw an exception if a resource with that id
  exist
* If prefetch fails, the Puppet run continues (this is fixed in 4.0)
* But the entire prefetch for that resource fails
* Resources won't correctly deal with exists?
* We could create extra resources or not delete others

This happens infrequently in the small, because it generally relies on
something else to be operating on the resources during the operation -
ie. between the initial describe_* call, another call deletes an
unrelated resource. But in the large that happens enough, especially
with the test suite.

This change swaps out the *_ids param for a filter on the *-id property.
This does exactly the same thing, but doesn't throw an exception. So now
we can simply deal with an empty array as needed.

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -46,10 +46,11 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
     end
     subnet_name = nil
     if instance.subnet_id
-      subnet_response = ec2_client(region).describe_subnets(subnet_ids: [instance.subnet_id])
-      subnet_name_tag = subnet_response.data.subnets.first.tags.detect { |tag| tag.key == 'Name' }
+      subnet_response = ec2_client(region).describe_subnets(filters: [
+        {name: 'subnet-id', values: [instance.subnet_id]}
+      ])
+      subnet_name = subnet_response.data.subnets.empty? ? nil : name_from_tag(subnet_response.data.subnets.first)
     end
-    subnet_name = subnet_name_tag ? subnet_name_tag.value : nil
 
     devices = instance.block_device_mappings.collect do |mapping|
       {
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -40,26 +40,31 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
         rule.user_id_group_pairs.collect do |security_group|
           name = security_group.group_name
           if name.nil?
-            group_response = client.describe_security_groups(
-              group_ids: [security_group.group_id]
-            )
-            name = group_response.data.security_groups.first.group_name
+            group_response = client.describe_security_groups(filters: [
+              {name: 'group-id', values: [security_group.group_id]}
+            ])
+            groups = group_response.data.security_groups
+            name = groups.empty? ? nil : groups.first.group_name
           end
-          {
-            'security_group' => name
-          }
+          if name
+            {
+              'security_group' => name
+            }
+          else
+            nil
+          end
         end
       end
-    end.flatten.uniq
+    end.flatten.uniq.compact
   end
 
   def self.security_group_to_hash(region, group)
     ec2 = ec2_client(region)
     vpc_name = nil
     if group.vpc_id
-      vpc_response = ec2.describe_vpcs(
-        vpc_ids: [group.vpc_id]
-      )
+      vpc_response = ec2.describe_vpcs(filters: [
+        {name: 'vpc-id', values: [group.vpc_id]}
+      ])
       vpc_name = if vpc_response.data.vpcs.empty?
         nil
       elsif vpc_response.data.vpcs.first.to_hash.keys.include?(:group_name)
--- a/lib/puppet/provider/ec2_autoscalinggroup/v2.rb
+++ b/lib/puppet/provider/ec2_autoscalinggroup/v2.rb
@@ -31,7 +31,9 @@ Puppet::Type.type(:ec2_autoscalinggroup).provide(:v2, :parent => PuppetX::Puppet
   def self.group_to_hash(region, group)
     subnet_names = []
     unless group.vpc_zone_identifier.nil? || group.vpc_zone_identifier.empty?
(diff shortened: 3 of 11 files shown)


lib/puppet/provider/ec2_securitygroup/v2.rb
```diff
@@ -40,26 +40,31 @@ def self.format_ingress_rules(client, group)
...
+          else
+            nil
           end
-          {
-            'security_group' => name
-          }
         end
       end
-    end.flatten.uniq
+    end.flatten.uniq.reject(&:nil?)
```
cmurphy: This could use the `.compact` method instead of `.reject(&:nil?)`

cmurphy:
Small nitpick, and this needs a rebase now.

garethr:
@cmurphy thanks, fixed and rebased. I'll convert a few other uses over to compact as well in my WIP PR.

zreichert:
Filed bug to track this 
CLOUD-254

cmurphy:
I keep seeing this during test runs:

```
An error occurred in an after(:all) hook.
  RSpec::Expectations::ExpectationNotMetError:
expected: false
     got: true

(compared using ==)
```

It's not actually causing a test failure, but it seems like it should.

I'm seeing a lot of other vpc-related failures as well, still trying to determine whether they're related to this change or just going over quota.

cmurphy:
The rspec expectation error is actually already appearing in master, and other failures seem to be related to quota issues or CLOUD-252.

Iristyle:
The tl; dr of this is.

* When using `describe_` APIs, passing it the relevant ids will throw an exception from the AWS side when those ids are invalid.

```
[~] aws ec2 describe-instances --instance-ids 'i-1234'                                                  15:55:11

A client error (InvalidInstanceID.NotFound) occurred when calling the DescribeInstances operation: The instance ID 'i-1234' does not exist
```

However, when using filters, that behavior changes to return an empty array.

```
[~] aws ec2 describe-instances --filters 'Name=instance-id,Values=i-1234'                               15:56:13
{
    "Reservations": []
}
```

Iristyle:
I've been thinking about this some more and talking to some of the Puppet core team -- seems like this could be a bad idea in general.... more details later, but I think we might want to revert these changes.

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -46,10 +46,11 @@ def self.instance_to_hash(region, instance)
     end
     subnet_name = nil
     if instance.subnet_id
-      subnet_response = ec2_client(region).describe_subnets(subnet_ids: [instance.subnet_id])
-      subnet_name_tag = subnet_response.data.subnets.first.tags.detect { |tag| tag.key == 'Name' }
+      subnet_response = ec2_client(region).describe_subnets(filters: [
+        {name: 'subnet-id', values: [instance.subnet_id]}
+      ])
+      subnet_name = subnet_response.data.subnets.empty? ? nil : name_from_tag(subnet_response.data.subnets.first)
```
Iristyle: I think the crux of my issue here is not correctly representing the underlying state of the resource, and then later making decisions based on this misrepresentation.  When code branches occur that are driven based on the incorrect state, this can lead to unpredictable behavior.

While it might seem to make sense that read-only properties on the resource are OK to omit based on transient failures, I think it actually makes the code much more difficult to reason about and later maintain. 

So I'm in favor of raising errors / failing fast when transient issues occur.  I would rather see the provider code fail as a result of AWS or other networking issues, rather than continue in a degraded state.

garethr: Until Puppet 4, if we raise exceptions here they get eaten by Puppet. The execution continues with no prefetch data, meaning even stranger things happen at a distance. I've some ideas about how we can better handle that but this change definitely raises less issues from usage in my experience.

garethr: It's also worth noting we've not seen any failures of that type in the acceptance tests (we generally assert on attributes coming back in the puppet resource tests.

What we did see were issues related to state in AWS being misleading. ie. AWS returning a resource id in an association that has already been deleted. This is obviously an eventual consistency thing on the AWS side which makes perfect sense.

Note that not all AWS resources have constrains around deleting connected resources, although some do.

This all means we will regularly receive IDs from AWS that don't exist, and failing in that condition isn't the correct thing to do. Even post-Puppet 4 failing in all these cases would mean regularly seeing entire runs fail due to the valid but inconsistent state of AWS.

Iristyle: I am familiar with the issue fixed in https://tickets.puppetlabs.com/browse/PUP-3656 / discussed on the PR at https://github.com/puppetlabs/puppet/pull/3428

I think there's a bit of confusion around `if we raise exceptions here they get eaten by Puppet`
  That's a bit imprecise, because that's only true for errors derived from `StandardError`.  We can simply create our own errors that derive directly from `Exception` and they will no longer be eaten in `prefetch`.

```ruby
class MyError < Exception
end

begin
  raise StandardError.new
rescue => foo
  puts "rescued: #{foo}"
end

begin
  raise MyError.new
rescue => bar
  puts "rescued: #{bar}"
end
```

```
[source] ruby test.rb                                                       10:27:28
rescued: StandardError
test.rb:11:in `<main>': MyError (MyError)
```

Regardless of where the state consistency issues are, I still don't believe it's a good idea to assume or otherwise act like that problem doesn't exist but papering over these details in `prefetch`

joshcooper: https://tickets.puppetlabs.com/browse/PUP-2643 is an example of what can go wrong when puppet swallows an exception during prefetch. And since PE creates a crontab entry (to run facter periodically) it has generated a PE support escalation. If the aws module fails to retrieve the current state, it should fail hard.

Iristyle:
So @garethr I'm going to revert this commit for now, because I've had enough conversations with core platform team that I feel like this approach has some issues.  I ran acceptance locally and I'm currently only seeing the single issue that's currently prevalent in CI as well.

Let's observe what happens in CI nightlies over the next few weeks and revisit.

Iristyle:
And normally I wouldn't have done this without catching up with you first, but I was working with @cmurphy on clearing up acceptance, and didn't want modifications there to confuse the revert (since it's also a change to prefetch) - see #116

PR #102, opened 2015-03-09
(CLOUD-42) Performance improvements

garethr:
This commit moves a bunch of functionality from the providers to the
base class. This centralisation allows for some basic caching.
Specically we regularly look up the name (often in a tag) of a resource
by it's AWS id.

This cuts down on the number of API calls made, especially as the
resource count increases.

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -6,6 +6,7 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
   confine feature: :retries
 
   mk_resource_methods
+  remove_method :tags=
 
   def self.instances
     regions.collect do |region|
@@ -67,6 +68,7 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
       instance_type: instance.instance_type,
       image_id: instance.image_id,
       instance_id: instance.instance_id,
+      id: instance.instance_id,
       monitoring: monitoring,
       key_name: instance.key_name,
       availability_zone: instance.placement.availability_zone,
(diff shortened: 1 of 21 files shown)


lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -63,6 +64,7 @@ def self.instance_to_hash(region, instance)
       instance_type: instance.instance_type,
       image_id: instance.image_id,
       instance_id: instance.instance_id,
+      id: instance.instance_id,
```
pjfoley: Is there a difference between this variable and the one above?

garethr: instance_id is exposed in the type, so you can access it via `puppet resource` if you need to. id is an implementation detail, it's used for shared methods (like the `tags=` method). So anything that want's to use `tags=` has to expose id in the `property_hash`.

lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -31,8 +35,16 @@ def self.read_only(*methods)
         end
       end
 
+      def self.logger
```
Iristyle: Could we pull the addition of logging out to a separate commit, since it doesn't really belong with the memoization?

joshcooper:
@garethr can you add the ticket id to the subject line for each commit, e.g. `(CLOUD-42) Memoize calls made to AWS by ID`, so it is possible to reconcile tickets and commits prior to release?

Iristyle:
:+1: to the git commit subject annotations - would like to see this across the board on all PRs.

pjfoley:
Out of curiosity where do you lodge tickets for this module?  I have looked at tickets.puppetlabs.com and cannot find a cloud-42 ticket?

garethr:
@pjfoley apologies for the confusion. We have an internal backlog to manage the user stories the team work from, but please open any issues you have here on GitHub.

Issue #100, opened 2015-03-07
Improve Route 53 support


danieldreier:
As far as I can tell, the Route 53 providers only cover A, NS, and TXT records. At a minimum, to manage the existing puppetlabs.com DNS using this, we'd also need:

- AAAA
- CNAME
- MX
- SPF

Beyond that, the ability to set Route 53 routing policies based on their simple, weighted, latency, failover, and geo-location options is important. We're not currently using any of those, but intend to fairly soon.

In particular, the ability to associate a DNS entry with a health check is significant, because for some services we're planning on running our own load balancers and would like to use DNS failover as part of that. I believe this would also require adding management of health checks to this module.

garethr:
@danieldreier thanks. We should be able to add the other DNS types quickly as that's mainly boilerplate.

We're definitely interested in adding routing options and health checks too, although that might take a little longer. If you have chance, an example manifest of how you imagine the health check type working would be particularly useful.

garethr:
This has been merged into master. Thanks for the suggestion.

PR #103, opened 2015-03-09
(CLOUD-257) Add support for AAAA, CNAME, MX, SRV and SPF records

garethr:
This addresses the extra resources requested by @danieldreier in https://github.com/puppetlabs/puppetlabs-aws/issues/100

### DIFF
--- a/lib/puppet/type/route53_spf_record.rb
+++ b/lib/puppet/type/route53_spf_record.rb
@@ -1,0 +1,15 @@
+require_relative '../../puppet_x/puppetlabs/route53_record'
+
+Puppet::Type.newtype(:route53_spf_record) do
+  extend PuppetX::Puppetlabs::Route53Record
+  @doc = 'Type representing a Route53 SPF record.'
+  create_properties_and_params()
+
+  # SPF records should always be wrapped in double quotes
+  # this munge avoids the need to pass in a '"value"' to Puppet
+  values_property = self.properties.find { |item| item == Puppet::Type::Route53_spf_record::Values }
+  values_property.munge do |value|
+    value =~ /^".+"$/ ? value : "\"#{value}\""
+  end
+end
+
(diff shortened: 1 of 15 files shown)


danieldreier:
This looks like what I'd need for now, and demonstrates the pattern of how to extend both the DNS types and test coverage. Thanks! +1

rhoml:
+1 this is a much needed feature

danieldreier:
Any chance we can add SRV records? We're using SRV to discover puppet masters.

garethr:
@danieldreier done, I've added SRV to this PR

danieldreier:
Awesome! Thanks!

lib/puppet/type/route53_spf_record.rb
```diff
@@ -0,0 +1,15 @@
...
+Puppet::Type.newtype(:route53_spf_record) do
+  extend PuppetX::Puppetlabs::Route53Record
+  @doc = 'Type representing a Route53 SPF record.'
+  create_properties_and_params()
+
+  # SPF records should always be wrapped in double quotes
+  # this munge avoids the need to pass in a '"value"' to Puppet
+  values_property = self.properties.find { |item| item == Puppet::Type::Route53_spf_record::Values }
+  values_property.munge do |value|
+    value =~ /^".+"$/ ? value : "\"#{value}\""
```
Iristyle: I think that we should pick one way of accepting these values and stick to it rather than allowing both.

garethr: The rationale here was that if you know DNS, then providing the quotes is more natural. If you don't know DNS it's the opposite. This also mirrors the behaviour of the AWS console.

My concern with choosing one is that without quotes is sort of wrong - the record will actually have quotes. But with quotes it's also unexpected if you don't know that. This felt like the nicest interface for different users.

danieldreier: I understand @Iristyle's concerns, but I think that having it just do the right thing is really valuable. In my personal experience, a fairly large proportion of SPF records are set blindly via copy-paste from some vendor's docs rather than being properly understood.

PR #104, opened 2015-03-10
(CLOUD-103) Allow mapping public ip addresses for instances launched in subnets

garethr:


examples/vpc-example/destroy.pp
```diff
@@ -1,3 +1,8 @@
+ec2_securitygroup { 'sample-sg':
+  ensure => absent,
+  region => 'sa-east-1',
+} ~>
+
```
cmurphy: This belongs in a separate commit.

cmurphy:
This parameter can't be changed after it's set. Puppet will report it as changing on every run without actually doing anything. It should be a read_only() method same as the others.

Iristyle:
Yes, agree with @cmurphy on this one - let's get those issues cleaned up and this should be pretty good to go.

garethr:
@cmurphy cheers. marked read only and moved the security group example out.

PR #115, opened 2015-03-17
Fix json syntax

rhoml:


garethr:
@rhoml thanks, much appreciated.

rhoml:
Your welcome. Linting was failing when updated.

PR #116, opened 2015-03-18
Filter on vpc-id for sec groups in EC2-VPC

cmurphy:
When adding security_group ingress rules to a security group resource
upon creation, a VPC-only account may have multiple groups with the
same name, often 'default', to select from. Without this change, the
ec2_securitygroup provider will simply select the first that it finds.
If the one it selects is not in the VPC to which the security group
puppet resource was added, it will raise an error and fail to add the
ingress rule. This change ensures that the correct security group for
the resource's ingress rule is selected by retrieving the VPC ID of the
security group that was added and using it as part of the filter for
selecting possible security groups to use for the ingress rule.

### DIFF
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -131,11 +131,15 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
     to_create.reject(&:nil?).each do |rule|
       if rule.key? 'security_group'
         source_group_name = rule['security_group']
-        group_response = ec2.describe_security_groups(filters: [
-          {name: 'group-name', values: [source_group_name]},
-        ])
+        filters = [ {name: 'group-name', values: [source_group_name]} ]
+        if vpc_only_account?
+          response = ec2.describe_security_groups(group_ids: [@property_hash[:id]])
+          vpc_id = response.data.security_groups.first.vpc_id
+          filters.push( {name: 'vpc-id', values: [vpc_id]} )
+        end
+        group_response = ec2.describe_security_groups(filters: filters)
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/ec2_securitygroup/v2.rb
```diff
@@ -131,11 +131,15 @@ def authorize_ingress(new_rules, existing_rules=[])
...
+        filters = [ {name: 'group-name', values: [source_group_name]} ]
+        if vpc_only_account?
+          response = ec2.describe_security_groups(group_ids: [@property_hash[:id]])
+          vpc_id = response.data.security_groups.first.vpc_id
+          filters.push( {name: 'vpc-id', values: [vpc_id]} )
+        end
+        group_response = ec2.describe_security_groups(filters: filters)
         fail("No groups found called #{source_group_name}") if group_response.data.security_groups.count == 0
-        source_group_id = group_response.data.security_groups.first.group_id
+        source_group_id = group_response.data.security_groups.find { |g| g.group_name == source_group_name }.group_id
```
Iristyle: Note that this was added based on what the AWS console shows when filtering - i.e. it does a substring match and returns `default_elb_XXX` when passing in `default` as the filter.

However, I've verified that in code / other environments, this is an exact string match as we've expected.

I modified code using the Ruby SDK to emit the list of security groups with `{:name=>"group-name", :values=>["default"]}` used as a filter, and only those that exactly match `default` are returned.

I also used the AWS CLI with `--filters "Name=group-name,Values=default"` to verify that only exact matches to `default` are returned.

So substring matched group names is not exactly the problem.  However, I do see two matches come back in our code -- but they're the same thing.  So I suspect something else is afoot.

Iristyle: Yes, I've confirmed that the only real issue here is that we weren't previously filtering on `vpc_id`.  Calling `find` shouldn't actually be necessary as was first assumed.

PR #117, opened 2015-03-18
(WIP)(CLOUD-184, CLOUD-205) Make the region param and ENV var interchangable

garethr:
(Note that I've put this up just applying to security groups so as to get feedback on whether we should do some/all/none this as much as the implementation)

This allows for creating and deleting resources from puppet resource by
passing the environment variable (which is already used for scoping).

Prior to this change you would have to pass the information twice. If
you do pass both we maintain the same approach as previous, giving the
param a higher precedence.

This also allows for not specifying the region in the DSL, and instead
using the AWS_REGION variable. This matches the standard behaviour of
other AWS tools, and means in the case of not providing anything the
correct error bubbles through.

### DIFF
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -24,7 +24,7 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
   def self.prefetch(resources)
     instances.each do |prov|
       if resource = resources[prov.name] # rubocop:disable Lint/AssignmentInCondition
-        resource.provider = prov if resource[:region] == prov.region
+        resource.provider = prov if (resource[:region] || ENV['AWS_REGION']) == prov.region
       end
     end
   end
@@ -96,14 +96,13 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
   end
 
   def exists?
-    dest_region = resource[:region] if resource
-    Puppet.info("Checking if security group #{name} exists in region #{dest_region || region}")
+    Puppet.info("Checking if security group #{name} exists in region #{target_region}")
     @property_hash[:ensure] == :present
   end
 
   def create
-    Puppet.info("Creating security group #{name} in region #{resource[:region]}")
-    ec2 = ec2_client(resource[:region])
+    Puppet.info("Creating security group #{name} in region #{target_region}")
+    ec2 = ec2_client(target_region)
     config = {
       group_name: name,
       description: resource[:description]
@@ -207,8 +206,8 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
   end
 
   def destroy
-    Puppet.info("Deleting security group #{name} in region #{resource[:region]}")
-    ec2_client(resource[:region]).delete_security_group(
+    Puppet.info("Deleting security group #{name} in region #{target_region}")
+    ec2_client(target_region).delete_security_group(
       group_id: @property_hash[:id]
     )
     @property_hash[:ensure] = :absent
--- a/lib/puppet_x/puppetlabs/aws.rb
+++ b/lib/puppet_x/puppetlabs/aws.rb
@@ -96,6 +96,12 @@ module PuppetX
         tags
       end
 
+      def target_region
+        target = resource ? resource[:region] || region : region
+        target = nil if target == :absent
+        target || ENV['AWS_REGION']
+      end
+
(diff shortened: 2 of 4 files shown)


lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -98,6 +98,12 @@ def self.tags_for(item)
         tags
       end
 
+      def target_region
+        target = resource ? resource[:region] || region : region
```
cmurphy: Where does `region` come from? Should this method have a parameter list `(region = default_region)`?

lib/puppet/provider/ec2_securitygroup/v2.rb
```diff
@@ -23,7 +23,7 @@ def self.instances
   def self.prefetch(resources)
     instances.each do |prov|
       if resource = resources[prov.name] # rubocop:disable Lint/AssignmentInCondition
-        resource.provider = prov if resource[:region] == prov.region
+        resource.provider = prov if (resource[:region] || ENV['AWS_REGION']) == prov.region
```
Iristyle: This is actually fairly tricky, and I'm not sure that this is the right approach.

Assume that:

* AWS has 2 security groups named `foo` in `us-east-1` and `us-west-1`.  
* In a manifest, you've defined a sg resource with `:region => 'us-west-1'`
* You've set `AWS_REGION=us-east-1`

It's now possible that even though the manifest wants to modify the resource in `us-west-1`, that Puppet will attach to and modify the resource in `us-east-1`

This may seem like an unlikely / synthesized problem, but it makes me really wary of doing something like this.  I'm not sure there's a reasonable way around this problem.

garethr:
Agreed. I'm unsure about this too.

I'm tempted to say if the env variable ever doesn't match the region value of a type we fail in the type validation. That blocks the entire run. Thoughts?

I think we should do that whether we merge this PR or not.

Iristyle:
Yes @garethr I was talking with @cmurphy and this seems like a tough nut to crack.  At first I suggested that we could use apply a rule that `AWS_REGION` env var takes over everything... but that's not a very good idea.

So your suggestion is to blow up the run when the resource defined in the manifest has a region that doesn't match the env var?  I think that's probably a good thing to do...

garethr:
@Iristyle yup. I like the filtering ability. It's a great performance optimisation, and it would allow for scaling out by region if you had lots of ec2 resources. But that power comes with complexity that I don't like.

If we stop the run at compile time if _any_ resource has a region that doesn't match the ENV then I think it's much safer. It means we never hit issues like the one you describe. We also catch issues we're you forget you have the ENV set.

lib/puppet_x/puppetlabs/aws.rb
garethr: region in this context is the value of the region property. This basically says:

* If you specify a region (ie. on the command line like `puppet resource ec2_securitygroup bob region=blah` or in the DSL, use that
* If that isn't present, fall back to the value of region for the resource

garethr:
@Iristyle @cmurphy I've changed this as discussed. Let me know if this works for you and I'll expand to the other relevant types.

lib/puppet/provider/ec2_securitygroup/v2.rb
zreichert: I don't like the idea of having two mechanisms that perform the same action.  Seems like it would get dirty very fast.  Which one would take president?  

[Here is a real world example](https://github.com/mrzarquon/tse_awsnodes/blob/master/manifests/vpc.pp) of modifying multiple resources with the same name in different regions. @mrzarquon mentioned that he wants to create identical VPCs in three different regions.

garethr:
I've extracted the region type and made a separate PR for just the blocking behaviour https://github.com/puppetlabs/puppetlabs-aws/pull/132

lib/puppet/provider/ec2_securitygroup/v2.rb
garethr: Regarding the two methods of passing region, this mirrors most of the AWS tools which will use the explicitly passed region OR fall back to the value of the Environment variable. It allows for some interesting reuse and in practice is less confusing that in theory in my experience. We should decide on whether we want to follow that convention, and if so agree we should better specify this behaviour and provide examples.

The linked example will only work if you have one puppet run per region, due to the requirement for unique resources in the graph. In the future we should definitely investigate other approaches here, including composite namevars ala https://github.com/puppetlabs/puppetlabs-aws/pull/124. A current pattern is to use the region var to make unique composite names. ie. {region}-{name}.

garethr:
I'm going to close this. It has a bunch of unanswered questions. It's hopefully useful as and when the core issue gets some time.

PR #118, opened 2015-03-18
(MAINT) Routes that are blackholed aren't active

garethr:
Avoid querying for gateways that don't exist. Once routes are
writable this could be improved.

### DIFF
--- a/lib/puppet/provider/ec2_vpc_routetable/v2.rb
+++ b/lib/puppet/provider/ec2_vpc_routetable/v2.rb
@@ -31,18 +31,20 @@ Puppet::Type.type(:ec2_vpc_routetable).provide(:v2, :parent => PuppetX::Puppetla
 
   def self.route_to_hash(region, route)
     ec2 = ec2_client(region)
-    gateway = if route.gateway_id == 'local'
-      'local'
-    else
-      begin
-        igw_response = ec2.describe_internet_gateways(internet_gateway_ids: [route.gateway_id])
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/ec2_vpc_routetable/v2.rb
```diff
@@ -31,18 +31,20 @@ def self.prefetch(resources)
...
   def self.route_to_hash(region, route)
     ec2 = ec2_client(region)
-    gateway = if route.gateway_id == 'local'
-      'local'
-    else
-      begin
-        igw_response = ec2.describe_internet_gateways(internet_gateway_ids: [route.gateway_id])
-        name_from_tag(igw_response.data.internet_gateways.first)
-      rescue Aws::EC2::Errors::InvalidInternetGatewayIDNotFound
+    if route.state == 'active'
```
Iristyle: This looks like something we can query for when making the call to `describe_route_tables`, rather than client side.

http://docs.aws.amazon.com/AWSEC2/latest/CommandLineReference/ApiReference-cmd-DescribeRouteTables.html

Any reason that we're not doing that?

garethr: We query for the route tables rather than explicitly for the routes. And although route tables support a filter for state it's misleading. Filtering for active will return all route tables (and routes for that table) with an active route. So a route table with one active and one black hole route will be returned and will have both routes, even if you filter at the route table level for active.

PR #120, opened 2015-03-18
(CLOUD-259) Add validation of types for properties

garethr:


cmurphy:
:+1:

cmurphy:
I might also add extra validation for things like IP addresses and subnets. However this ticket is just about type validation so we can discuss whether additional validation is needed independent of this.

PR #121, opened 2015-03-18
(MAINT) Remove the additional query for VPC lookup for security groups

garethr:
Security groups can only associate with those in the same VPC. See
the following gist for an example of what happens when you try:

https://gist.github.com/garethr/81ef20bba61e89c037e6

That means it's only valid for the security group we're connecting with
to have the same VPC id as the security group we're changing. So by
stashing the vpc_id locally we can avoid the extra query to look it up.

This also accomodates VPCs in Classic accounts.

### DIFF
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -77,6 +77,7 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
       ensure: :present,
       ingress: format_ingress_rules(ec2, group),
       vpc: vpc_name,
+      vpc_id: group.vpc_id,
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/ec2_securitygroup/v2.rb
```diff
@@ -105,6 +106,8 @@ def create
       vpc_id = vpc_response.data.vpcs.first.vpc_id
```
Iristyle: I think there was one issue here that you were going to address - `vpc_name` doesn't have to be set when you're creating a new `sg` in a VPC only account.  So under those circumstances, (none specified but VPC only account) we have to query for the `vpc_id` that has been attached.

garethr: Added and pushed.

PR #55, opened 2015-01-20
(CLOUD-70) Handle much more complex security groups

garethr:
Previously we had support for quite simple security group types. This
issue gives an idea of a more realistic target:
https://github.com/puppetlabs/puppetlabs-aws/issues/42

This work in progress code aims to handle that issue, as well as
multiple variations. I have local test cases that need turning into acceptance
tests, and some more of the rules code can be pulled into the new class.

The class itself also wants a bit of tidying up and independent tests.
But at the moment this works for me locally against a non-trivial
security group config so woot.

Issue #42, opened 2015-01-06
Error handling on security_group creation


rhoml:
Hi @garethr we already have our cloud in AWS up and running so I am not provisioning any security group which we provision using piculet.

When I add a security group that already exists into puppetlabs-aws I get errors like:
`Error: The permission 'x.x.x.x/32-1-22-22' has already been authorized on the specified group` 

Is that an expected behaviour or perhaps something that should be improved?.

garethr:
Could you post an example of the manifest you're using? That's not the expected behaviour, what you're doing should work.

rhoml:
It looks like this

 ``
 ec2_securitygroup { 'security-group-name':
    ensure           => present,
    region           => $::ec2_placement_region,
    description      => 'XXXXXXX',
    ingress          => [{
      protocol       => 'tcp',
      port           => 6379,
      security_group => 'other-secgroup'
    },
    {
      protocol       => 'udp',
      from_port      => 0,
      to_port        => 65535,
      security_group => 'other-secgroup'
    },
    {
      protocol       => 'tcp',
      port           => 22,
      cidr           => 'x.x.x.x/32'
    }
    ],
    tags        => {
      tag_name => 'value',
    },
  }
``

garethr:
Ah, so we don't quite support that setup yet, although we should do. And we're not validating the ingress rule structure very well yet either.

The details are in
https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/provider/ec2_securitygroup/v2.rb#L91-L110

Basically you can specify a security group OR a protocol/port/cidr combination at present.

So you could do:

```puppet
 ec2_securitygroup { 'security-group-name':
    ensure           => present,
    region           => $::ec2_placement_region,
    description      => 'XXXXXXX',
    ingress          => [{
      security_group => 'other-secgroup'
    },{
      protocol       => 'tcp',
      port           => 22,
      cidr           => 'x.x.x.x/32'
    }],
    tags        => {
      tag_name => 'value',
    },
  }
```

This is nearly, but not quite, what you're after (it opens up tcp ports other than 6379 to the other security group).

I'll leave this issue open here and we'll improve the flexibility and validation of the ingress rules in an upcoming release. Thanks for the example, the more real world cases we have the better the module will be.

rhoml:
Oh, I see. Well sounds ok to me.

ahenroid:
[...]

PR #122, opened 2015-03-18
id security groups primarily by name (falling back to vpc_name / group_name) and much more

keymone:
this is proof of concept, it is also a mix of 2 large changes (sorry)

1. ec2_securitygroup name param value is written into both group_name and tags['Name'] and is used to uniquely identify the group

2. more complete implementation of ingress rules

cc @garethr

### DIFF
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -18,7 +18,7 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
     end.flatten
   end
 
-  read_only(:region, :description)
+  read_only(:name, :region, :description)
 
   def self.prefetch(resources)
     instances.each do |prov|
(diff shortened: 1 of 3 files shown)


garethr:
Hi @keymone, thanks for looking a this. It's probably worth pairing this back to one thing at a time I'm afraid.

It's also worth tracking https://github.com/puppetlabs/puppetlabs-aws/pull/55, which contains support for much more complex security group rules (see the acceptance tests and unit tests for examples of those), it supports the suggestions in https://github.com/puppetlabs/puppetlabs-aws/issues/42 and more.

garethr:
@keymone The code in https://github.com/puppetlabs/puppetlabs-aws/pull/124 supports setting rules on default security groups in VPC using a composite namevar. Let me know if that works for you if you have a moment.

garethr:
Given the overlap with #55, and the work in #124 and the lack of movement on this I'm moved to close the PR. Apologies for not finding more time to push this particular feature forward.

PR #123, opened 2015-03-18
a more complete impl for ingress rules

keymone:
port range, -1 for protocol, multiple cidrs or security groups

```
ingress     => [ { protocol => '-1' }, # all traffic, source - current group
                 { protocol => '-1',  cidr => '10.0.0.0/8' }, # all traffic, cidr
                 { protocol => 'tcp', port => 10 }, # tcp port 10, current group
                 { protocol => 'tcp', port => 12, security_group => 'test' }, # ref group by name
                 { protocol => 'tcp', port => [10, 2000] } # port range
]
```

@garethr i have problems with running this against acceptance tests, getting this error:

```
ec2_securitygroup
  should create a new security group
Notice: Compiled catalog for -.com in environment production in 0.06 seconds
Error: Could not set 'present' on ensure: missing region; use :region option or export region name to ENV['AWS_REGION'] in 1
Error: Could not set 'present' on ensure: missing region; use :region option or export region name to ENV['AWS_REGION'] in 1
Wrapped exception:
missing region; use :region option or export region name to ENV['AWS_REGION']
Error: /Stage[main]/Main/Ec2_securitygroup[-@--5-efa4aea3-b4b6-4899-aff7-dac9e604e9cc]/ensure: change from absent to present failed: Could not set 'present' on ensure: missing region; use :region option or export region name to ENV['AWS_REGION'] in 1
Notice: Finished catalog run in 6.29 seconds
    with the specified name
    isn't attached to a VPC (EC2-Classic accounts only)
    with the specified tags
    with the specified description
    with the specified ingress rules (FAILED - 1)
```


garethr:
Could you take a look at https://github.com/puppetlabs/puppetlabs-aws/pull/55 for reference. These appear to support the same thing to this PR but with slightly different interfaces.?

The changes there are backwards compatible I think, and support the examples in https://github.com/puppetlabs/puppetlabs-aws/issues/42.

If I'm missing something let me know. It would probably be useful to consolidate on #55 if possible.

keymone:
@garethr i did check the #55, there are few things i don't like:

- iterating over tcp/upd/icmp when protocol=-1 is a valid value
- unnecessary explicit expansion of from_ and to_ ports when nils are valid values
- it's more code and feels more complex (this is subjective ofc)

i can try landing my impl onto #55 and see what happens but not sure it makes sense when this PR is twice smaller. i'll try fixing the acceptance tests (they don't run cleanly against #55 either btw) and see where it gets us.

garethr:
@keymone unfortunately -1 isn't a valid value in non-VPC accounts: http://docs.aws.amazon.com/sdkforruby/api/Aws/EC2/Client.html#authorize_security_group_ingress-instance_method.

Not sure I follow on the expansion point? My view is it would be useful to support explicit `from_port` and `to_port` options which mirror the API conventions. Where those two values are the same you can use the shorthand `port`.

As a design rules we've been keeping things quite close to the API, for those people that already know it. So I've a preference for `from_port` and `to_port` over `port => [1,2]`.

I'll take a closer look at the code when I get a moment, their will undoubtedly be shorter ways of doing what I'm doing at present.

keymone:
interesting, both here - http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_AuthorizeSecurityGroupEgress.html and here - http://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html there is no mention that value -1 is invalid to non-VPC accounts..

as for ports - i agree that keeping close to API makes sense but i also think API should be minimal - a rule AWS doesn't seem to follow at all. for me handling single `port` param was better than handling three params `from_port`, `to_port` and `port`. i'm not going to insist on this though.

the expansion point means that if neither `from_` nor `to_` ports are specified we shouldn't be telling aws to set them to 1 and 65535 which is what #55 does here(?): https://github.com/puppetlabs/puppetlabs-aws/pull/55/files#diff-7f9b57f89472b6db8594a41706bdcbfaR137

keymone:
@garethr  regarding protocol=-1 - would using `#vpc_only_account?` to decide whether we want to send -1 or 3 ip_permissions for each protocol make sense?

garethr:
Just tidying up PRs. I'm going to close this as #55 was merged and supports most of this, as discussed with a slightly different interface.

PR #124, opened 2015-03-19
Allow setting ingress rules for default security groups in VPC

garethr:
Due to default security groups all being named default we couldn't
reference them previously due to unique resouce naming conflicts. This
patch allows for a composite namevar only in the case of the default
group. Note that the composite name populates the VPC field
automatically, so you don't have to duplicate the information in a
separate property.


garethr:
Note that this still requires some acceptance tests to verify the functionality before merging.

This relates to a conversation I had a few days ago with @keymone.

keymone:
@garethr i had to change title patterns to make it work:

```
def self.title_patterns
    [ [ /^(([\w\-]+)::(default))$/,
        [ [ :name, lambda {|x| x} ],
          [ :vpc, lambda {|x| x} ],
          [ :group_name, lambda {|x| x} ] ] ],
      [ /^((.*))$/,
        [ [ :name, lambda {|x| x} ],
          [ :group_name, lambda {|x| x} ] ] ] ]
  end
```

seems like it's not possible to specify custom group_name with this implementation? like:

```
ec2_securitygroup { "long unique name across regions":
  group_name: "shorter per-region unique name"
}
```

or am i missing some obvious way to do that?

garethr:
@keymone Ah, that's probably better to be a bit more liberal for the second capture. Thanks, will change.

Correct on the second formulation not working with this PR. Reversing this becomes much more difficult so I stuck to just what you can't do at the moment - ie. default security groups in VPCs.

keymone:
@garethr is there any progress with this? i can't access https://tickets.puppetlabs.com/browse/CLOUD-269 to check. we're using this patch for a while now and it's working just fine.

garethr:
@keymone unfortunately no progress to report, mainly due to other prioritise I'm afraid. I'll update this issue whenever things do move.

ceturc:
This looks like excellent work.  I have multiple VPCs and this fix will really help me manage my security groups with Puppet.  Right now, during an apply, puppet just picks the first one it finds with a matching name - kind of dangerous!   I look forward to seeing this pulled into the published forge module.

ccaum:
Any update on this? Otherwise I'm going to close due to age.

PR #128, opened 2015-03-20
Add function to return the content of a file from an s3 bucket

pjfoley:
I would like to have the ability to pull a file from an s3 bucket and manage it via the built in puppet file type resource.  I have raised a puppet ticket asking for this to be considered - https://tickets.puppetlabs.com/browse/PUP-4080

I created a PR to the puppet code based to add this capability to puppet core (https://github.com/puppetlabs/puppet/pull/3675)

Quite rightly it was suggested this capability be added to the puppetlabs-aws module, the only issue is there is no mechanism to allow a module to extend the core puppet file type to pull content from another source.

As a temporary mechanism until it is possible to extend the puppet file type, I have created a function that will download a file from aws s3 and return the raw content to the file type.  Essentially a user would use this function like so:

```puppet
file { '/tmp/example.txt':
        content => s3_file_content('foo-bucket', 'bar-file', 'ap-southeast-2'),
	replace => false,
}
```

One thing that I am not sure if it is a puppet bug, a limitation in the puppet language or by design - the s3_file_content function is called every puppet run regardless if the file exists or not.  Which means it will make a call to s3 (which does cost money each 1000 api calls) Thoughts?

### DIFF
--- a/lib/puppet/parser/functions/s3_file_content.rb
+++ b/lib/puppet/parser/functions/s3_file_content.rb
@@ -1,0 +1,35 @@
+require_relative '../../../puppet_x/puppetlabs/aws.rb'
+
+module Puppet::Parser::Functions
+  newfunction(:s3_file_content, :type => :rvalue, :doc => <<-ENDHEREDOC
+              Downloads a file from aws s3 and returns the contents
+
+              Requires three parameters:
+                Bucket
+                File
+                Region
+
+              ENDHEREDOC
+             ) do |args|
+
+     unless args.length == 3 then
+       raise Puppet::ParseError, ("s3_file_content(): wrong number of arguments (#{args.length}; must be 3)")
+     end
+
+     bucket = args[0]
+     filename = args[1]
+     region = args[2]
+
+     ["bucket", "filename", "region"].each do |param_name|
+       var = eval(param_name)
+        unless var.instance_of?(String) then
+          raise Puppet::ParseError, ("Parameter [#{param_name}] is not a string.  It looks to be a #{var.class}")
+        end
+        if var.to_s == "" then
+          raise Puppet::ParseError, ("Parameter [#{param_name}] cannot be blank")
+        end
+     end
+
+     PuppetX::Puppetlabs::Aws.s3_client(region).get_object( bucket: bucket, key: filename).body.read
+   end
+end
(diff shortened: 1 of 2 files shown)


danieldreier:
@pjfoley I've been considering taking a stab at something similar, but my idea had been to create a new type (filesync?) that implements a more generic bi-directional concept, with something like:

```
# on a CA master:
filesync { '/etc/puppet/ssl/crl.pem':
  provider => 's3',
  source   => 's3://example_bucket/crl.pem',
  sync       => 'push',
}

# on a compile master:
filesync { '/etc/puppet/ssl/crl.pem':
  provider => 's3',
  source   => 's3://example_bucket/crl.pem',
  sync       => 'pull',
  mode      => '0644',
  user        => 'puppet',
  group      => 'puppet',
}
```

However, since this could be implemented using s3, rsync, ftp, etc my plan had been to implement it as a standalone module. I believe that making the request using either `If-Modified-Since` or `If-Match` on the ETag would be the only options for verifying if a change had been made. If we were willing to cache data somewhere we could store the `x-amz-version-id` response to see if the object had changed, but that would be a bit awkward to store locally between runs.

Let me know if that's something you'd be interested in collaborating on as a standalone module.

pjfoley:
@danieldreier sorry about the delay in responding for some reason I missed seeing this response.

I am time poor at the moment but might be interested in looking into this.  However have you seen this [1] puppet ticket.  I think it would be better to wait until they have built in a way to do this.  That way we can just store information in the file bucket natively.  Hopefully there will also be away to extend the metadata stored as well which would make it super easy.

I would suggest voting on the issue and leaving a comment mentioning the other file sources you are interested in.  The more votes the more likely it will be worked on.

[1] - https://tickets.puppetlabs.com/browse/PUP-4368

danieldreier:
@pjfoley thanks for pointing me to that. I've left an encouraging comment.

garethr:
I'm just tidying up PRs and I'm going to close this unless anyone objects. It feels like the conversation moved on to whether this should be in Puppet or a separate provider which I think sounds like the best option.

pjfoley:
Personally I feel that this is the correct location for the code.  If you are using AWS you will also be using S3 and it makes sense to have this sort of capability for users of the module.

To the best of my knowledge the changes have never been made to puppet to enable providing providers via modules.  @kylog has the status of this changed at all?

Refs:
https://github.com/puppetlabs/puppet/pull/3675
https://tickets.puppetlabs.com/browse/PUP-4368

But it is your call.

PR #129, opened 2015-03-20
Readme init edit

jbondpdx:
More light edits to come, but this is pretty heavily added to / revised.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,8 @@
 
 1. [Overview](#overview)
 2. [Description - What the module does and why it is useful](#module-description)
-3. [Setup - Getting started](#setup)
+3. [Setup - Requirements](#setup)
+4. [Getting Started](#getting-started)
 4. [Usage - Configuration options and additional functionality](#usage)
 5. [Reference - An under-the-hood peek at what the module is doing and how](#reference)
 5. [Limitations - OS compatibility, etc.](#limitations)
@@ -25,6 +26,15 @@ importantly, it allows you to describe your entire AWS infrastructure and
 to model the relationships between different components.
 
 ##Setup
+
+###Requirements
+
+* Puppet 3.4 or greater
+* Ruby 1.9
+* Amazon AWS Ruby SDK (available as a gem)
+* Retries gem
+
+###Installing the aws module
 
 1. Install the retries gem and the Amazon AWS Ruby SDK gem. 
 
@@ -81,11 +91,30 @@ region using an environment variable.
 export AWS_REGION=eu-west-1
 ```
 
+## Getting Started with aws
+
+To stand up an instance with AWS, use the `ec2_instance` type:
+
+~~~
+ec2_instance { ['web-1', 'web-2']:
+  ensure          => present,
+  image_id        => 'ami-67a60d7a', # EU 'ami-b8c41ccf',
+  security_groups => ['web-sg'],
+  instance_type   => 't1.micro',
+  tags            => {
+    department => 'engineering',
+    project    => 'cloud',
+    created_by => $::id,
+  }
+}
+~~~
+
+
 ##Usage
 
 ### Creating and Destroying Resources 
 
-#### Using the Puppet DSL
+#### Create and Destroy with the Puppet DSL
 
 The aws module allows you to manage AWS using the Puppet DSL. Let's create a simple stack, with a load balancer, instances, and security groups.
 
@@ -133,7 +162,7 @@ To destroy the resources created by the above, run the following:
 puppet apply tests/destroy.pp --test
 ```
 
-#### From the Command Line
+#### Create and Destroy on the Command Line
 
 The module also has basic `puppet resource` support, so you can manage AWS resources from the command line. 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,14 +91,34 @@ region using an environment variable.
 export AWS_REGION=eu-west-1
 ```
 
+## Getting Started with aws
+
+To stand up an instance with AWS, use the `ec2_instance` type:
```
garethr: this example actually creates 2 instances. It's also missing the region property, and requires a security group. The image_ids are also somewhat specific to the example it's taken from. The tagging is also very Puppet Labs specific. The deleted example is much better in my view.

README.md
```diff
@@ -165,78 +184,566 @@ and then destroy them, all from the command line:
...
+#####`vpc`
+*Optional* VPC to assign the route table to.
+
+#####`region`
+*Optional* The region in which to launch the route table. For valid values, see [AWS Regions](http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region).
+
+#####`routes`
+*Optional* Individual routes for the routing table. Accepts an array of 'destination_cidr_block' and 'gateway' values:
+ 
+~~~
```
garethr: any reason not to puppet highlight these with ```puppet?

README.md
```diff
@@ -165,78 +184,566 @@ and then destroy them, all from the command line:
 puppet resource ec2_securitygroup test-group ensure=absent region=sa-east-1
 ```
 
+### Managing AWS Infrastructure
+
+You can use the aws module to audit AWS resources, launch autoscaling groups in VPC, perform unit testing, and more. The [examples](examples/) directory contains a variety of usage examples that should give you an idea of what's possible:
+
+* [Puppet Enterprise](puppet-enterprise/): Start up a small Puppet Enterprise cluster using the AWS module.
```
garethr: the links here won't work as they're relative. You'll need to add the examples prefix to the URL

README.md
```diff
@@ -18,64 +19,64 @@ Status](https://travis-ci.org/puppetlabs/puppetlabs-aws.svg?branch=master)](http
...
+###Requirements
 
-    gem install aws-sdk-core
+* Puppet 3.4 or greater
+* Ruby 1.9
+* Amazon AWS Ruby SDK (available as a gem)
+* Retries gem
 
-We also use the retries library so install that with:
+###Installing the aws module
```
garethr: Some titles appear to be sentence case while others are leading caps. We should probably be consistent.

README.md
```diff
@@ -165,78 +184,566 @@ and then destroy them, all from the command line:
...
-  tags              => {
-    tag_name => 'value',
-  },
-}
-```
+### Types
 
-### ec2_securitygroup
-
-```puppet
```
garethr: I think the examples give an idea of what the modules is about much better than the list of examples or the details of the types. I'd be loath to lose them from the README. In fact I wonder about them being nearer the top (ie before the install instructions)

README.md
```diff
@@ -165,78 +184,566 @@ and then destroy them, all from the command line:
...
-  monitoring        => true,
-  key_name          => 'name-of-existing-key',
-  security_groups   => ['name-of-security-group'],
-  user_data         => template('module/file-path.sh.erb'),
-  tags              => {
-    tag_name => 'value',
-  },
-}
-```
+### Types
```
garethr: This section is so long it feels like at best it needs an index. I wonder if this would be better as separate files in a directory, with links here in the README.

README.md
jbondpdx: Yeah, I knew the example needed some attention and likely had. I sent you an email on Thursday with questions about it. I can just go back to the earlier example.

README.md
jbondpdx: Good question! Docs has been moving to ~~~ as it renders properly in our docs; my understanding is that the language specific highlights do not. I think there are good reasons to use the language fences, but they may not work for us everywhere. I'll have some conversations about this tomorrow.

README.md
jbondpdx: The Reference sections can be terribly long, I know. We try to avoid nesting in readmes, and I think users are used to having the Reference info in one document. That said, I think what would help a lot is adding subheads under Reference in the TOC, and I will do that.

README.md
jbondpdx: I am definitely OK with adding the security group and loadbalancer examples back into the Usage section. The installation/setup section is traditionally at the top in all module readmes; we try not to change things too much, as users are accustomed to it.

README.md
jbondpdx: Indeed we should! Good catch, thank you. H1 and H2 are leading caps, the rest should be sentence case.

README.md
jbondpdx: So, I followed up on this today, and the reason is that ```puppet doesn't render properly in our docs. As long as a doc never leaves Github, the language highlighting fencing is fine, but we are working with the assumption that Readmes may eventually be pulled into our docs site.

README.md
jbondpdx: Rather than add subheads in the TOC, I've added a list of types first, followed by a list of parameters by type. This is consistent with other long readmes we have, and should let the user see all the types without having to wade through all the params at the same time.

jbondpdx:
@garethr This PR has been updated, if you could please take another look at it. Thank you very much!

README.md
```diff
@@ -86,18 +97,96 @@ determining if something is available. This can be a little slow. If you
...
-```
+~~~
+
+##Getting Started with aws
+
+The aws module allows you to manage AWS using the Puppet DSL. To stand up an instance with AWS, use the `ec2_instance` type. The following code sets up a very basic instance: 
+
+~~~
+ec2_instance { 'instance-name':
+  region          => 'us-west-1',
```
garethr: Incredibly picky, could you reformat like so. This meets the style guide by having ensure first and having the right amount of whitespace between the property names and the arrows. 

~~~
ec2_instance { 'instance-name':
  ensure        => present,
  region        => 'us-west-1',
  image_id      => 'ami-123456', # you need to select your own AMI
  instance_type => 't1.micro',
}
~~~

PR #130, opened 2015-03-23
(WIP, CLOUD-179) Terminate the run if we get inconsistent data from AWS

garethr:
If we query AWS for a resource (by id) that doesn't exist we will raise
an exception. Due to PUP-3656 that prevents prefetch but lets the
provider continue with incorrect data.

This changes that behaviour to stop the provider run if we discover
inconsistent data, it also provides information to help debug the
problem - although in many cases running a second time will fix the
issue.

Common causes of inconnsistency are:

* what appears to be eventual consistency within AWS
* other clients acting on the API at the same time


garethr:
Note that this was done around `self.instances` rather than `prefetch` as listing resources using puppet resource doesn't actually touch `prefetch` and uses `self.instances` directly.

PR #132, opened 2015-03-24
(CLOUD-184, CLOUD-205) Avoid ambiguity when using the AWS_REGION environment var

garethr:
If using the AWS_REGION environment variable and a manifest where the
region property of one or more resources doesn't match you can see
unspecified behaviour. See
https://github.com/puppetlabs/puppetlabs-aws/pull/117 for discussions.

This PR extracted from the above simply stops that from happening. If
the env is present and doesn't match All of the region properties then
we fail. This maintains the utility of the env var as an optimisation
but stops us shooting ourselves in the foot.


DavidS:
You have missed the region property in lib/puppet/type/ec2_instance.rb.

ahenroid:
I updated and posted https://github.com/puppetlabs/puppetlabs-aws/pull/401

PR #138, opened 2015-03-30
(MAINT) Add a fairly advanced example using future parser and stdlib

garethr:


examples/distribute-across-az/init.pp
```diff
@@ -0,0 +1,74 @@
...
+$number_of_app_servers = 3
+$maximum_app_servers = 10
+$region = 'sa-east-1'
+$ami = 'ami-67a60d7a'
+
+# Don't change below this line
+# First lets validate the variables passed in
+validate_array($availability_zones)
+validate_string($region)
+validate_string($ami)
```
danieldreier: Since we're already using future parser, we could use assert_type:
```puppet
assert_type(Array[String[1]], $availability_zones)
assert_type(String[1], $region)
assert_type(String[1], $ami)
```

The String[1] validates that it's a non-empty string.

garethr: Ah, nice. Thanks. I've also updated the following checks using the Integer type.

examples/distribute-across-az/init.pp
```diff
@@ -0,0 +1,71 @@
...
+  }
+}
+
+# Finally lets delete any nodes we no longer require
+# this allows you to reduce the $number_of_app_servers
+# variable and delete any left-over instances
+range($number_of_app_servers+1, $maximum_app_servers).each |$node| {
+  $instance_name = "app-${node}"
+  ec2_instance { $instance_name:
+    ensure => absent,
```
danieldreier: this is clever - good idea. I'll be taking this back to the code on the ops side. Had been scratching my head about how to do that.

danieldreier:
+1

examples/distribute-across-az/init.pp
garethr: The other option is the `resources` metatype: https://docs.puppetlabs.com/references/latest/type.html#resources. You could set that to purge and noop for reporting, but that assumes management of all resources. This is much more constrained (I need to put that in an example now I think about it)

examples/distribute-across-az/init.pp
```diff
@@ -0,0 +1,71 @@
...
+# of instances or launch them in different availability zones
+$availability_zones = ['sa-east-1a', 'sa-east-1b']
+$number_of_app_servers = "5"
+$maximum_app_servers = 10
+$region = 'sa-east-1'
+$ami = 'ami-67a60d7a'
+
+# Don't change below this line
+# First lets validate the variables passed in
+assert_type(Array[String[1]], $availability_zones)
```
DavidS: The whole lower part could be wrapped into a define. Then the `assert_type` calls here could be replaced by type annotations on the parameters of the define.

DavidS:
Except for the subjective comment, :+1:

Issue #144, opened 2015-04-07
Request limit exceeded


stepanstipl:
Hi, when trying to use this module and create single ec2 instance, I'll mostly (like 9 in 10 tries) end up with **`Request limit exceeded.`** error (I'll get `/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/lib/ruby/2.0.0/delegate.rb:295: [BUG] Segmentation fault
ruby 2.0.0p481 (2014-05-08 revision 45883) [universal.x86_64-darwin14]` in the rest of the tries). The AWS environment has over 1k instances, so it's not exactly small, but I've been using 'aws-sdk-core' directly without any issues in this environment. Looks like module is making too many API calls and hitting EC2 api requests limit, let me know what would help you to debug this issue... I'm looking into it myself.

I'm running puppet like this:
````
puppet apply init.pp --hiera_config ./hiera.yaml --modulepath ./modules --debug
````

Versions used:
- aws-sdk-core: 2.0.28
- puppet: 3.7.4
- puppetlabs-aws: 1.0.0
- ruby: 2.0.0p481

Thanks, Stepan

````
...
Info: Applying configuration version '1428425803'
Debug: Prefetching v2 resources for ec2_instance
Debug: Storing state
Debug: Stored state in 0.00 seconds
Debug: Using settings: adding file resource 'rrddir': 'File[/Users/stepan/.puppet/var/rrd]{:path=>"/Users/stepan/.puppet/var/rrd", :mode=>"750", :ensure=>:directory, :loglevel=>:debug, :links=>:follow, :backup=>false}'
Debug: Finishing transaction 70179195852760
Debug: Received report to process from stepan-mb.local
Debug: Processing report from stepan-mb.local with processor Puppet::Reports::Store
Error: Could not run: Puppet detected a problem with the information returned from AWS
when looking up ec2_instance in eu-west-1. The specific error was:

Request limit exceeded.

Rather than report on ec2_instance resources in an inconsistent state we have exited.
This could be because some other process is modifying AWS at the same time.
````

stepanstipl:
So it seems like the problem is in provider `instance_to_hash`, which is called for every instance when puppet's doing prefetch. As we have many instances this will be going on for a while and eventually end up with `Request limit exceeded`.  I guess better way would be to prefetch all subnets in one call and use that.

garethr:
Hi @stepanstipl, thanks for the details.

[...]

PR #146, opened 2015-04-08
Prefetch subnets info in one call - quite big performance boost

stepanstipl:
This helped me with issue #144 when when i was hitting `Request limit exceeded` error after  couple of minutes (our env is quite big - 1k+ instances in given region). Now it seems to work as expected and finish in about 10 seconds in this environment (I was trying just to list resources with `puppet resource ec2_instance` command).

This should be properly tested though as I don't have environment to do so. Currently 3 VCR test will fail (1 of those used to fail for me even before my change), which I guess is expected, and I'm unable to run beaker tests currently (don't have AWS environment ready for it).

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -11,12 +11,22 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
     regions.collect do |region|
       begin
         instances = []
+        subnets = Hash.new()
+
+        subnets_response = ec2_client(region).describe_subnets()
+        subnets_response.data.subnets.each do |subnet|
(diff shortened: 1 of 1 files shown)


garethr:
@stepanstipl thanks for this example, that's perfect. I've got a pretty good test environment setup for this so I'm more than happy to fix up the tests and the like. I'm travelling the next few weeks but I'll update this issue when I get time to implement this. Thanks again.

garethr:
Hi @stepanstipl. Just a quick update. I've taken this pattern and run with it in https://github.com/puppetlabs/puppetlabs-aws/pull/102, which also contains a number of other improvements. I'll close this PR in favour of that, but thanks again for testing this out and suggesting the change.

stepanstipl:
Cool, great, I'll have a look and give #102 a go and see how it looks for us, cheers

PR #147, opened 2015-04-08
(Cloud-297) ec2_instances should support tenancy

mrzarquon:
adds the instance_tenancy parameter to ec2_instance, enabling individual
nodes to be added to dedicated hardware as needed. The only way to
accomplish this previously was thru creating a VPC where this was the
default behavior instead.

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -258,6 +258,7 @@ Found #{matching_groups.length}:
         instance_initiated_shutdown_behavior: resource[:instance_initiated_shutdown_behavior].to_s,
         placement: {
           availability_zone: resource[:availability_zone]
+          tenancy: resource[:instance_tenancy]
         },
         monitoring: {
           enabled: resource[:monitoring].to_s,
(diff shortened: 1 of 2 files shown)


lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -258,6 +258,7 @@ def create
         instance_initiated_shutdown_behavior: resource[:instance_initiated_shutdown_behavior].to_s,
         placement: {
           availability_zone: resource[:availability_zone]
```
garethr: looks like this might be missing a trailing `,`.

lib/puppet/type/ec2_instance.rb
```diff
@@ -114,6 +114,12 @@ def insync?(is)
     end
   end
 
+  newproperty(:instance_tenancy) do
```
garethr: I think this could probably just be tenancy, given it's on the instance type?

garethr:
This only provides the option to create the instance with a specified tenancy, it doesn't populate the hash with this value in `self.instance_to_hash`. Also:

Is this read_only after creation of the instance? If so it should be added to: https://github.com/mrzarquon/puppetlabs-aws/blob/CLOUD_297/lib/puppet/provider/ec2_instance/v2.rb#L31

We'll need to add basic unit and acceptance tests as well, and update the README types docs. Let me know if you want to do that or I can pick it up later.

tphoney:
I am closing this in preference of the PR #295 which was merged

PR #149, opened 2015-04-13
Added ssl_certificate upload

s4mur4i:
WIP, added ssl_certificate upload,
- do I also need to add the sub to the create function to invoke it at creation time?

Have a support case to validate how it can be removed if the entry is removed.


DavidS:
Hi, will you be working on this any more?

s4mur4i:
Hy,

Haven't done any work on it in time. I think it can be closed.

DavidS:
:-(

PR #156, opened 2015-04-28
Link fix

jbondpdx:
Fixed some links that were relative to GitHub, so, sadly, didn't work on the Forge. Now they work from both.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,18 +255,18 @@ puppet resource ec2_securitygroup test-group ensure=absent region=sa-east-1
 
 ### Managing AWS infrastructure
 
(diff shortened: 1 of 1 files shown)


garethr:
Mmmm. This breaks links on any forks (ie. they no longer point to that fork) and breaks links when viewed locally. This feels like something that the Forge should resolve, although we might need to merge to work around until that happens. Let's chat with the Forge folks before merging.

jbondpdx:
Well, phooey. I should have known it was too easy a fix.

garethr:
As discussed, we decided this is the less of two (minor) evils so I'm going to go ahead and merge.

PR #161, opened 2015-05-06
(CLOUD-295) - Add RDS support

petems:
Continued from #69 

Main requirements for merge from last commits: More spec tests


lib/puppet/provider/rds_db_parameter_group/v2.rb
```diff
@@ -0,0 +1,37 @@
+require_relative '../../../puppet_z/puppetlabs/aws.rb'
```
garethr: should be `puppet_x` rather than `puppet_z`

lib/puppet/type/rds_db_parameter_group.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  newparam(:name, namevar: true) do
+    desc 'the name of the DB Parameter Group (also known as the db_parameter_group_name)'
+  end
+
+  newproperty(:description) do
+    desc 'the description of a DB parameter group'
+  end
+
+  newproperty(:db_parameter_group_family) do
```
garethr: Would this make sense just as family? Given it's already in the context of a db_parameter_group

lib/puppet/type/rds_db_parameter_group.rb
```diff
@@ -0,0 +1,25 @@
...
+    desc 'the name of the DB parameter group family that this DB parameter group is compatible with.'
+  end
+
+  newproperty(:region) do
+    desc 'the region in which to create the db_parametergroup'
```
garethr: missing an underscore in the db_parametergroup

lib/puppet/type/rds_db_securitygroup.rb
```diff
@@ -0,0 +1,36 @@
+Puppet::Type.newtype(:rds_db_securitygroup) do
+  @doc = 'Type representing an RDS instance.'
+
+  ensurable
+
+  newparam(:name, namevar: true) do
+    desc 'the name of the DB Security Group (also known as the db_security_group_name)'
+  end
+
+  newparam(:db_security_group_description) do
```
garethr: should just be description

lib/puppet/type/rds_db_securitygroup.rb
```diff
@@ -0,0 +1,36 @@
...
+    validate do |value|
+      fail 'db_security_group_description should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:owner_id) do
+    desc 'the ID of the owner of this DB Security Group'
+  end
+
+  newproperty(:ec2_security_groups, :array_matching => :all) do
```
garethr: Given this property, should this type have an autorequire on `ec2_securitygroup`?

petems: Hmm, I thought I'd already fixed that, must've been lost in a push somewhere...

lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -0,0 +1,122 @@
...
+      license_model: instance.license_model,
+      multi_az: instance.multi_az,
+      iops: instance.iops,
+      backup_retention_period: backup_retention_period,
+      skip_final_snapshot: skip_final_snapshot,
+      final_db_snapshot_identifier: final_db_snapshot_identifier,
+      db_parameter_group_name: instance.db_parameter_groups.collect(&:db_parameter_group_name).first,
+      db_security_groups: instance.db_security_groups.collect(&:db_security_group_name),
+    }
+    if instance.respond_to?('endpoint')
```
garethr: This doesn't check for endpoint being nil. I just hit in this in the after part of the acceptance tests.

lib/puppet/provider/rds_db_parameter_group/v2.rb
petems: Fixed

lib/puppet/provider/rds_instance/v2.rb
petems: Wait, it can respond to endpoint but the endpoint can actually be nil? Like when it's just started up?

petems: Would

```ruby
if instance.respond_to?('endpoint') && !instance.endpoint.nil?
``` 

be ok to fix this? Or is there a smarter way?

garethr: That should work. From the testing I did I didn't see it _not_ respond to endpoint, But starting it up and then shutting it down quickly could result in endpoint being nil.

petems: Ok, will change to that and add a comment with that :+1:

garethr:
Closing in favour of https://github.com/puppetlabs/puppetlabs-aws/pull/165

PR #163, opened 2015-05-11
(CLOUD-306) Allow specifiying an http proxy for all API traffic

garethr:


DavidS:
I've tested this locally and it worked as advertised.

Issue #160, opened 2015-05-04
ec2_elastic_ip fails in autoscaling (same name)


CarmenAPuccio:
https://github.com/puppetlabs/puppetlabs-aws/blob/master/examples/elastic-ip-addresses/README.md
ec2_elastic_ip needs to support instance_id as the same name can be used multiple times which leads to failure. Also, in an autoscale scenario it doesn't attach to the new node because the terminated node still carries the name you are trying to attach to. If it was instance_id you could use factor facts for ec2_instance_id in your manifests. Please help and support instance_id as well. Error below.

Warning: Multiple instances found named ePixFTPQAv1.0, using i-6cbf8743
Error: Could not set 'attached' on ensure: stopped waiting, encountered a failure state at 21:/opt/releases/puppet/manifests/site.pp
Error: Could not set 'attached' on ensure: stopped waiting, encountered a failure state at 21:/opt/releases/puppet/manifests/site.pp
Wrapped exception:
stopped waiting, encountered a failure state
Error: /Stage[main]/Main/Ec2_elastic_ip[52.7.57.238]/ensure: change from detached to attached failed: Could not set 'attached' on ensure: stopped waiting, encountered a failure state at 21:/opt/releases/puppet/manifests/site.pp

ec2_elastic_ip { '52.7.57.238':
  ensure   => 'attached',
  instance => 'ePixFTPQAv1.0',
  region   => 'us-east-1',
}

ahenroid:
@CarmenAPuccio, reading through the thread, I believe this was resolved for you by #188. Can I close this issue? Thanks!

PR #164, opened 2015-05-12
ec2_elastic_ip instance_id support bug fix for autoscaling.

CarmenAPuccio:
Changing the c2_elastic_ip module to support instance_id because if two
ec2 instances exist with the same name like in AutoScaling the module
fails because of duplicate names. Now with instance_id you can target a
specific instance along with the name. This is referenced in issue:
ec2_elastic_ip fails in autoscaling (same name) #160

### DIFF
--- a/examples/elastic-ip-addresses/README.md
+++ b/examples/elastic-ip-addresses/README.md
@@ -28,12 +28,28 @@ a different account, and IP addresses are unique.
 
 With the module installed as described in the README, from this directory run:
 
-    puppet apply init.pp
+	puppet apply ec2_create.pp
 
-This creates the instances and associates the IP address to the one
-called `web-1`. We can see that by running:
+This creates the EC2 instances called web-1 and web-2.
 
-    puppet resource ec2_elastic_ip
+From here you need to get the Instance ID of the EC2 instance.
+The easiest ways of doing that are from the AWS console, or from the
+command line tools like so:
+
+	aws ec2 describe-instances --filters "Name=tag:Name,Values=web-1"
+
+	aws ec2 describe-instances --filters "Name=tag:Name,Values=web-2"
+
+Once you have the Instance ID's of the two web instances you need to modify the manifest in `init.pp`.
+This is because the Instance ID present in the file is an example and ID's are unique to your account.
+
+With the module installed as described in the README, from this directory run:
+
+	puppet apply init.pp
+
+This associates the IP address to the one called web-1, We can see that by running:
+
+	puppet resource ec2_elastic_ip
 
 Which should return something like:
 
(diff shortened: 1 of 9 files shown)


DavidS:
Hi,

this looks basically OK to me, but needs squashing and rebasing on top of the current version, as it cannot be merged automatically.

Please see these docs for help with the git:
Rebasing: https://git-scm.com/book/en/v2/Git-Branching-Rebasing

Regards, David

CarmenAPuccio:
Squash and rebase done. Let me know if you need anything else.

examples/elastic-ip-addresses/README.md
```diff
@@ -45,6 +61,11 @@ ec2_elastic_ip { '177.71.189.57':
 }
 ~~~
 
-We can now use `puppet resource` to switch the IP to the `web-2` instance:
+We can now use `puppet resource` to switch the IP to the `web-2` instance (Don't forget to change the IP and instance_id):
+
+<<<<<<< HEAD
```
garethr: Looks like code left in from a merge

examples/elastic-ip-addresses/README.md
```diff
@@ -28,12 +28,28 @@ a different account, and IP addresses are unique.
...
-This creates the instances and associates the IP address to the one
-called `web-1`. We can see that by running:
+This creates the EC2 instances called web-1 and web-2.
 
-    puppet resource ec2_elastic_ip
+From here you need to get the Instance ID of the EC2 instance.
+The easiest ways of doing that are from the AWS console, or from the
+command line tools like so:
+
+	aws ec2 describe-instances --filters "Name=tag:Name,Values=web-1"
```
garethr: I'd prefer to use puppet resource to get the instance_id if we need it.

lib/puppet/provider/ec2_elastic_ip/v2.rb
```diff
@@ -61,10 +61,18 @@ def create
     instance_ids = response.reservations.map(&:instances).flatten.map(&:instance_id)
 
     fail "No instance found named #{resource[:instance]}" if instance_ids.empty?
-    if instance_ids.count > 1
-      Puppet.warning "Multiple instances found named #{resource[:instance]}, using #{instance_ids.first}"
+
+    if instance_ids.include?(resource[:instance_id])
+      Puppet.info("Found an instance with a name of #{resource[:instance]} and matches the id of #{resource[:instance_id]}")
+      instance_ids.delete_if {|x| x != resource[:instance_id]}
+    else
```
garethr: could use `elsif` here

lib/puppet/type/ec2_elastic_ip.rb
```diff
@@ -35,6 +35,14 @@
     end
   end
 
+  newproperty(:instance_id) do
+    desc 'The id of the instance associated with the Elastic IP'
+    validate do |value|
+      fail 'instance_id should be a String' unless value.is_a?(String)
+      fail 'You must provide an instance_id for the Elastic IP association' if value.nil? || value.empty?
```
garethr: I don't think this is the correct approach, or at least isn't in keeping with the module's current design. Using the ID in extremis maybe, but requiring it as the primary interface I'm less sure about I'm afraid.

garethr:
I think the overall issue is sane, ie. how best to describe the use of elastic IPs for autoscaling instances. I don't however think the approach works. Partly because it bleeds through the instance_id, which could change as the instances are scaled up or down, and partly because we've worked to avoid doing so elsewhere.

CarmenAPuccio:
I'm closing this because the more I look at #188 the more it fits my need anyway. I've already tested this for my use case and it works fine. Because terminated instances linger around for a bit you could run into a conflict but now that #188 looks for only pending and started, you should never get a failure because of the terminated instance with the same name when trying to attach the EIP.

Issue #137, opened 2015-03-29
Provider ec2_vpc and ec2_vpc_routetable - Issue managing route tables


pjfoley:
Hi,

Using the vpc-example provided in the repository I have been testing the VPC provider.  Although the puppet code works it does surprising things which I would not expect.  If I run:

```puppet
ec2_vpc { 'sample-vpc':
  ensure       => present,
  region       => 'sa-east-1',
  cidr_block   => '10.0.0.0/16',
}
```

This creates a VPC **and** a route table both named **sample-vpc**.  Looking through the aws-sdk-core manual it does not explicitly say it will create a route table but I assume this is correct behavior.

If it is correct behavior I would like to use this route table (sample-vpc) and add an additional route to it instead of creating a separate route table. For example (**does not work**):

```puppet
ec2_vpc { 'sample-vpc':
  ensure       => present,
  region       => 'sa-east-1',
  cidr_block   => '10.0.0.0/16',
}

ec2_vpc_internet_gateway { 'sample-igw':
  ensure => present,
  region => 'sa-east-1',
  vpc    => 'sample-vpc',
}

#################
# The below will fail as there is already a sample-vpc route table created as part of the ec2_vpc type
#################
ec2_vpc_routetable { 'sample-vpc':
  ensure => present,
  region => 'sa-east-1',
  vpc    => 'sample-vpc',
  routes => [
    {
      destination_cidr_block => '10.0.0.0/16',
      gateway                => 'local'
    },{
      destination_cidr_block => '0.0.0.0/0',
      gateway                => 'sample-igw'
    },
  ],
}
```

However the routes property for the ec2_vpc_routetable type is read only once created and I am unable to add an additional route to an existing route table.

garethr:
So, the route table you mentioned is automatically created by AWS, rather than by Puppet or the SDK. We tag it with the same name as the VPC to make keeping track of it possible.

However, as you mentioned, it's not (yet) possible to update the routes on an existing route table. This isn't a hard limitation, it's definitely something we'll be adding the ability to do. At which point the code you have should just work.

daveseff:
I had the same issue. Error message below. Strangely, All my work is being don in us-west-1. Don't know why it's complaining about us-east-1:

Error: Could not run: Puppet detected a problem with the information returned from AWS
when looking up ec2_vpc_routetable in us-east-1. The specific error was:

expected params[:internet_gateway_ids][0] to be a string

[...]

PR #171, opened 2015-05-19
Update README per CLOUD-214

psoloway:
* Add a note to the Reference section for each parameter that is
read-only after creation.
* Remove the CI Build Status as part of a site-wide shift to reduce
server load.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,3 @@
-[![Puppet
-Forge](http://img.shields.io/puppetforge/v/puppetlabs/aws.svg)](https://forge.puppetlabs.com/puppetlabs/aws) [![Build
-Status](https://travis-ci.org/puppetlabs/puppetlabs-aws.svg?branch=master)](https://travis-ci.org/puppetlabs/puppetlabs-aws)
-
 ####Table of Contents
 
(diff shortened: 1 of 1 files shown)


pjfoley:
Some of these parameters should be changeable after creation.  Is this just to reflect the way the types and providers currently work?

For example #137

garethr:
@pjfoley yes, you're correct. This documentation change is just updating the docs to represent how things currently work.

garethr:
Hi @psoloway. The main README changes look good.

If you could just change the commit message. Ideally it should follow the format:

    (CLOUD-XXX) details of why the change was made

The backlog isn't public and it's useful to have the why if you're going through the commit log.

I'm definitely not in favour of removing the travis indicator however. The commit mentions this in passing as reducing load, but the load here is on Travis. This visibility is very important to anyone involved in the development. If you could remove that from this commit at least and maybe grab me to discuss.

garethr:
Closed in https://github.com/puppetlabs/puppetlabs-aws/commit/7c9bfae473e5f4a74ea426c923a440558d472e80 after talking with @jbondpdx

PR #175, opened 2015-06-08
(MAINT) We shouldn't care about the return order of availability zones

garethr:
This test just started failing in Jenkins. The test is overly
prescriptive in that we want both availability zones returned but we
don't care about the order. On the other hand this has to my knowledge
never failed before so it's fixed order, just whatever is fixing it has
changed.

This change makes the test more robust, simply allowing both possible
options.

### DIFF
--- a/spec/acceptance/loadbalancer_spec.rb
+++ b/spec/acceptance/loadbalancer_spec.rb
@@ -160,7 +160,7 @@ describe "ec2_loadbalancer" do
         end
 
         it 'availablity_zones' do
-          regex = /(availability_zones)(\s*)(=>)(\s*)(\[\'sa\-east\-1a\', \'sa\-east\-1b\'\])/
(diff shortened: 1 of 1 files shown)


DavidS:
The more 'correct' regexp would be

`/availability_zones\s*=>\s*\[(\'sa-east-1a\', \'sa-east-1b\'|\'sa-east-1b\', \'sa-east-1a\')\]/`

  * remove unnecessary grouping (also existing on many other regexes in the file, maybe not worth to fix here)
  * remove `\-` escapes
  * do not match `-a, -a` or `-b, -b`

garethr:
@DavidS cheers, I've replaced with the suggested regex. I'll merge once Travis confirms nothing is amis so as to get the pipeline building.

bmjen:
+1 TravisCI is green, will merge this to run it through jenkins.

PR #177, opened 2015-06-10
(WIP) (FM-2802) Introduce global config and credentials files in Puppet confdir

garethr:
This is to work around the Puppet agent nuking environment variables,
meaning the default location (based on HOME) doesn't work correctly.
This also adds support for storing the other settings in global files.

### DIFF
--- a/lib/puppet_x/puppetlabs/aws.rb
+++ b/lib/puppet_x/puppetlabs/aws.rb
@@ -36,7 +36,7 @@ This could be because some other process is modifying AWS at the same time."""
       end
 
       def self.default_region
-        ENV['AWS_REGION'] || 'eu-west-1'
+        ENV['AWS_REGION'] || region_from_global_configuration || 'eu-west-1'
       end
 
       def default_region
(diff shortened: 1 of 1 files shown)


daveseff:
#thumbsup

esalberg:
We forked this change, and it's working for us, so merging it in would be great.

aharden:
Consider this another vote for getting this merged in ASAP.  I've been running off a fork of this module for months to be able to use our http proxy for AWS API access during Puppet agent runs.

lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -56,21 +56,85 @@ def self.read_only(*methods)
...
 
       def self.client_config(region)
-        config = {region: region, logger: logger}
-        if ENV['PUPPET_AWS_PROXY'] and not ENV['PUPPET_AWS_PROXY'].empty?
-          config[:http_proxy] = ENV['PUPPET_AWS_PROXY']
+        config = {logger: logger}
+        config[:http_proxy] = proxy_configuration if proxy_configuration
+        config[:credentials] = global_credentials if global_credentials
+        config[:region] = if global_configuration
+          config[:region] = region_from_global_configuration || region
```
DavidS: either put the assignment into the if, or the if into the assignment. Please, not both :-)

lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -56,21 +56,85 @@ def self.read_only(*methods)
...
+        else
+          region
         end
         config
       end
 
+      # This method is vendored from the AWS SDK, rather than including an
+      # extra library just to parse an ini file
```
DavidS: Just to make sure: this is not part of the aws-sdk-core gem this currently requires, and adding this as a runtime dependency would require all users to install additional code on nodes running the module?

DavidS:
I only had two minor comments on the code style. Other than that, this should amend the current README. Recommending the IAM role assignment harder and pointing out that `~/.aws` only works with `puppet apply` would be a good start.

DavidS:
While this is still mergable, if you go in an modify the commit, a rebase would also clean up the history.

DavidS:
Merged with slight modifications as #246 .

Issue #152, opened 2015-04-22
Referencing ephemeral instance storage in block_devices mapping


ginkel:
Is there a way to reference ephemeral instance storage for the block_devices mapping of an ec2_instance? Reading the source I'm under the impression that only EBS may be supported...

benpillet:
I'm running into the same issue.  I tried specifying the virtual_name as ephemeral0, but still no luck.  Terraform splits out the root_block_device, ebs_block_device, and ephemeral_block_device into separate attributes as noted here: https://www.terraform.io/docs/providers/aws/r/instance.html.  I could see the implementation going either way.  To keep backwards compatibility, we could add another field to the array entry for a block_device with a name like storage_type and values like ['ebs', 'instance'] with ebs being the default.  Or split out the different storage types like terraform.  I was thinking of working on this in the next few days.  Any preference on the implementation?

ahenroid:
@ginkel @benpillet, I believe this addressed by #212. Can I close this issue?

benpillet:
Yes.  Looks like #212 covers it.  Thanks!

ahenroid:
Thanks, @benpillet!

PR #182, opened 2015-06-12
Configure ephemeral/instance store block devices too

benpillet:
As a first step for Issue #152 I still need to add documentation.  Feedback would be appreciated for whether I'm on the right track or not.  Thanks :-)


DavidS:
Hi @benpillet , that doesn't look half bad! README and test fixes would be mandatory, though :-)

See also #212 which is modifying exactly the same place for snapshots. @cpick and you might want to coordinate on this.

mentat:
Ping.  Any changes you can add some tests and README updates?

PR #186, opened 2015-06-23
(WIP- DO NOT MERGE) refactor tests to use Beaker

zreichert:
These tests should now work with beaker and without beaker.  
To run without beaker:
DEVELOPMENT=true bundle exec rspec spec/acceptance 
To run with beaker 
bundle exec rspec spec/acceptance 

All beaker-rspec env vars work.  The tests assume that the agent used to do the provisioning will have the default role.  A default nodeset is provided with this PR.

### DIFF
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -1,6 +1,24 @@
 require 'aws-sdk-core'
 require 'mustache'
 require 'open3'
+unless ENV['DEVELOPMENT']
+  require 'beaker-rspec'
+  require 'beaker/puppet_install_helper'
+  install_pe
+  on(default, puppet('module install puppetlabs-aws'))
+
+  # install the AWS SKD & retries gems on the agent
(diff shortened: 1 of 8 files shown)


spec/spec_helper_acceptance.rb
```diff
@@ -282,13 +284,59 @@ def self.puppet_resource(resource, opts = {}, command_flags = '')
...
-    response = shell(cmd)
+    response = Toggle_switch.shell(cmd)
     response
   end
 
 end
 
-class TestExecutor::Response
+# a common class to use to switch between local shell access and an Agent as SUT
+class Toggle_switch
```
garethr: CamelCase class names rather than _ as per Ruby style

garethr:
I'm closing this in favour of https://github.com/puppetlabs/puppetlabs-aws/pull/186, which is based on the initial commit but just fixing up the rest of the tests due to the interface change.

PR #190, opened 2015-07-01
Fix issue with routes with nil gateway_id

mikeslattery:
Routes that have don't have a gateway id were causing the following error:
   expected params[:internet_gateway_ids][0] to be a string


mikeslattery:
For anyone interested, here is an alternative workaround:

In your module create lib/puppet/provider/ec2_vpc_routetable/v2_patch.pp:

    Puppet::Type.type(:ec2_vpc_routetable).provide(:v2_patch, :parent => :v2) do
      def self.gateway_name_from_id(region, gateway_id)
        gateway_id ? super.gateway_name_from_id(region, gateway_id) : nil
      end
    end

Put in any manifest with route tables:

    Ec2_vpc_routetable {
        provider => v2_patch,
    }

rfletcher:
Related: #192

garethr:
Thanks for fixing. I've squashed the commits together and merged into master. Much appreciated. https://github.com/puppetlabs/puppetlabs-aws/commit/74d3fd19ce92b8d1b4dca9c027b51cd0670e2b9b

PR #193, opened 2015-07-14
Update IAM profile.json example

claflico:
Adds missing IAM permissions necessary for listing ELBs and adding internet gateways.


garethr:
@claflico thanks for spotting those and submitting the PR. Much appreciated.

PR #198, opened 2015-07-21
(CLOUD-376) Update installation instructions with new paths

garethr:


README.md
```diff
@@ -47,19 +47,19 @@ to model the relationships between different components.
...
 1. Install the retries gem and the Amazon AWS Ruby SDK gem.
 
-    * If you're using open source Puppet, the SDK gem should be installed into the same Ruby used by Puppet. Install the gems with these commands:
+  * If you're running Puppet from the `puppet-agent` package or running Puppet Enterprise 2015.2.0 you can install the gems using:
 
-      `gem install aws-sdk-core`
+      `/opt/puppetlabs/puppet/bin/gem install aws-sdk-core retries`
 
-      `gem install retries`
+  * If you're using an older version of the open source release of Puppet the gems should be installed into the same Ruby used by Puppet. Install the gems with these commands:
```
jbondpdx: Some of this is nitpicking, but I'd change this sentence to something more like: 

"If you're using an older version of open source Puppet, install the gems into the same Ruby used by Puppet with these commands: "

Some of this is nitpicking, but I'd change this sentence to something more like: 

"If you're using an older version of open source Puppet, install the gems into the same Ruby used by Puppet with these commands: "

"The gems should be installed" leaves me unsure as to whether you are telling me to do something. I think the missing comma might be an American/British difference.

PR #201, opened 2015-07-29
(CLOUD-410) Fix issue with managing RDS instances in subnets

garethr:


danieldreier:
I tested using the following manifest and got the behavior I expected. It was created in the correct VPC / db subnet group.

```puppet
rds_instance { "pl-db-dev":
  ensure               => present,
  allocated_storage    => '20',
  db_instance_class    => 'db.t2.small',
  db_parameter_group   => 'website',
  db_subnet            => 'default-vpc-5adf103f',
  master_user_password => 'fakepassword1',
  engine               => 'mysql',
  engine_version       => '5.6.22',
  license_model        => 'general-public-license',
  master_username      => 'root',
  multi_az             => false,
  region               => 'us-west-2',
  storage_type         => 'gp2',
}
```

spec/acceptance/rds_spec.rb
```diff
@@ -92,6 +92,12 @@ def get_rds_instance(name)
       end
     end
 
+    it 'with the default subnet association (VPC-only accounts)' do
+      unless @aws.vpc_only?
+        expect(@rds_instance.db_subnet_group.db_subnet_group_name).to eq('default')
```
DavidS: Are there any expectations connected to the non-`vpc_only` case?

If yes, they should be added, if no, can we perhaps push this information to a tag or add a `pending` or `skip` as alternative to make this more clear? Also connected: is this constant throughout a test run?

garethr: I've reformulated this as a single test rather than two tests. The test above contained the reverse expectations.

DavidS: I'm still not very fond of the `@aws.vpc_only?` dependency, but I do not understand the problem enough to suggest a better alternative. Would merge.

PR #202, opened 2015-08-04
Update profile.json in README.md, fix json syntax errors

vazhnov:
Fix this errors from https://console.aws.amazon.com/iam/home#roles:

If I use example from profile.json:
«This policy contains the following JSON error on line 16: Expected ',' instead of '"'»

If I use example from README.md:
«This policy contains the following JSON error on line 83: Bad string»


DavidS:
Thanks!

PR #203, opened 2015-08-06
Ignore ELB instance order

rfletcher:
`Elb_loadbalancer` should ignore the order of instances for comparison. 

After creating an ELB with a definition like this...:

````puppet
elb_loadbalancer { 'web':
  ...
  instances => [
    'web-1',
    'web-2',
  ],
  ...
}
````

...subsequent runs failed without modification to the manifest, because the AWS API returned the instances in a different order:

    Notice: /Stage[main]/Elb_loadbalancer[web]/instances: current_value ["web-2", "web-1"],
    should be web-1 web-2

This change ignores the array's order for comparison, matching behavior of the `subnets` and `security_groups` parameters.


DavidS:
Thanks!

PR #205, opened 2015-08-12
Route53: Always use upsert

bashtoni:
Intermittently, for reasons I didn't manage to track down, route53 A record updates were failing because the module believed it should do a create instead.

I solved this by using UPSERT rather than CREATE, allowing update of existing records.

Example debug output from the problem below:

```
Info: Checking if A record kibana.nonprod.redacted.co.uk. exists
Info: Creating A record kibana.nonprod.redacted.co.uk.
Error: Could not set 'present' on ensure: Tried to create resource record set [name='kibana.nonprod.redacted.co.uk.', type='A'] but it already exists at 29:/etc/puppetlabs/code/localmodules/r53hostname/manifests/init.pp
Error: Could not set 'present' on ensure: Tried to create resource record set [name='kibana.nonprod.redacted.co.uk.', type='A'] but it already exists at 29:/etc/puppetlabs/code/localmodules/r53hostname/manifests/init.pp
Wrapped exception:
Tried to create resource record set [name='kibana.nonprod.redacted.co.uk.', type='A'] but it already exists
Error: /Stage[main]/R53hostname/Route53_a_record[kibana.nonprod.redacted.co.uk.]/ensure: change from absent to present failed: Could not set 'present' on ensure: Tried to create resource record set [name='kibana.nonprod.redacted.co.uk.', type='A'] but it already exists at 29:/etc/puppetlabs/code/localmodules/r53hostname/manifests/init.pp
```

### DIFF
--- a/lib/puppet/provider/route53_record.rb
+++ b/lib/puppet/provider/route53_record.rb
@@ -77,7 +77,7 @@ class Puppet::Provider::Route53Record < PuppetX::Puppetlabs::Aws
   def create
     Puppet.info("Creating #{self.class.record_type} record #{name}")
     route53_client.change_resource_record_sets(
-      record_hash('CREATE')
+      record_hash('UPSERT')
     )
     @property_hash[:ensure] = :present
   end


DavidS:
Thanks!

I guess this could be a race condition, if you are running the same manifest on multiple agents or something else is poking your route53.

DoctahPopp871:
This is the best. I have been going after this one for the past week!

PR #206, opened 2015-08-13
(CLOUD-412, CLOUD-411) Support specifying backup retention

garethr:
It was pointed out that the AWS console defaults the backup retention to
30 days, however the API defaults it to 1 day. Because we we're exposing
that parameter this was both impossible to change and a bit invisible.

This change adds the ability to manage the backup retention period for
new RDS instances.

In the process of testing that change to ensure it was idempotent I
added the relevant (missing) test. This passed for the new property but
failed for another one (allocated_storage). This was reported separately
in CLOUD-411. This commit fixes that too.

### DIFF
--- a/lib/puppet/type/rds_instance.rb
+++ b/lib/puppet/type/rds_instance.rb
@@ -95,9 +95,11 @@ Not applicable. Must be null.'
 
   newproperty(:allocated_storage) do
     desc 'The size of the DB.'
+    munge do |value|
+      value.to_i
+    end
     validate do |value|
-      fail 'allocated_storage type should not contains spaces' if value =~ /\s/
-      fail 'allocated_storage should not be blank' if value == ''
+      fail 'allocated_storage must be an integer' unless value.to_i.to_s == value.to_s
     end
   end
 
@@ -193,6 +195,17 @@ Not applicable. Must be null.'
     validate do |value|
       fail 'final_db_snapshot_identifier should be a String' unless value.is_a?(String)
       fail 'final_db_snapshot_identifier should not be blank' if value == ''
+    end
+  end
+
+  newproperty(:backup_retention_period) do
+    desc 'The number of days to retain backups. Defaults to 30 days.'
+    defaultto 30
+    munge do |value|
+      value.to_i
+    end
+    validate do |value|
+      fail 'backup_retention_period must be an integer' unless value.to_i.to_s == value.to_s
     end
   end
 
--- a/spec/unit/type/rds_instance_spec.rb
+++ b/spec/unit/type/rds_instance_spec.rb
@@ -33,6 +33,7 @@ describe type_class do
       :endpoint,
       :port,
       :db_parameter_group,
+      :backup_retention_period,
     ]
   end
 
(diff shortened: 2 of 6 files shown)


lib/puppet/type/rds_instance.rb
```diff
@@ -196,6 +198,17 @@ def insync?(is)
     end
   end
 
+  newproperty(:backup_retention_period) do
+    desc 'The number of days backups should be retained for.'
```
DavidS: Please use the same wording as the README. "The number of days to retain backups. Defaults to 30 days." just reads much better.

spec/unit/type/rds_instance_spec.rb
```diff
@@ -108,4 +109,41 @@
...
+
+  it 'should be happy with strings for backup_retention_period' do
+    machine = type_class.new(:name => 'sample', :backup_retention_period => 40)
+    expect(machine.property(:backup_retention_period).insync?('40')).to be true
+  end
+
+  it 'should be happy with integers for backup_retention_period' do
+    machine = type_class.new(:name => 'sample', :backup_retention_period => 45)
+    expect(machine.property(:backup_retention_period).insync?(45)).to be true
+  end
```
DavidS: This and the prior test case could be joined into a common context:

```
  context 'with the backup_retention_period set to an integer' do
    let(:machine) { type_class.new(:name => 'sample', :backup_retention_period => 40) }

    it 'should be happy with strings for backup_retention_period' do
      expect(machine.property(:backup_retention_period).insync?('40')).to be true
    end

    it 'should be happy with integers for backup_retention_period' do
      expect(machine.property(:backup_retention_period).insync?(40)).to be true
    end
  end
```

The same could be done for all the default test cases, and probably for a number of the happy-path insync tests to share a single instance.

spec/unit/type/rds_instance_spec.rb
```diff
@@ -108,4 +109,41 @@
...
+
+  it 'should default backup_retention_period to 30' do
+    srv = type_class.new(:name => 'sample')
+    expect(srv[:backup_retention_period]).to eq(30)
+  end
+
+  it 'allocation storage must be an integer' do
```
DavidS: please use the full property name in the message to aid searching and parsing

spec/unit/type/rds_instance_spec.rb
```diff
@@ -108,4 +109,41 @@
     end
   end
 
+  it 'backup retention must be an integer' do
```
DavidS: please use the full property name in the message to aid searching and parsing

DavidS:
This looks good to me, except for some aesthetics.

danieldreier:
looks good to me; I think @DavidS's feedback is reasonable.

lib/puppet/type/rds_instance.rb
garethr: Fixed.

spec/unit/type/rds_instance_spec.rb
garethr: I've fixed up the new tests. I've left the others due to size, although now the test suite as a hole is some-what stable it would be nice to go through a refactor when time permits.

spec/unit/type/rds_instance_spec.rb
garethr: Fixed.

spec/unit/type/rds_instance_spec.rb
garethr: Fixed.

PR #208, opened 2015-08-13
Corrects curl PE tarball download URL

aharden:
Apparently direct download via S3.amazonaws.com was disabled at some point.  I found that this works.


DavidS:
Thanks!

PR #211, opened 2015-08-18
[BUGFIX] rds_instance with VPC db_subnet

Remz-Jay:
Trying to create an `rds_instance` with the `db_subnet` parameter failed with:
```
Error: Could not set 'present' on ensure: unexpected value at params[:subnet_group_name]
```
The AWS CLI manual for `rds create-db-instance` specifies the parameter as `[--db-subnet-group-name <value>]`.

Updating the parameter to the correct name fixes the error and results in
correct creation of an RDS instance within the VPC subnet group.


garethr:
@Remz-Jay thanks for looking at this. I already have a fix up for this in https://github.com/puppetlabs/puppetlabs-aws/pull/201. That PR also contains a few other related fixes and acceptance tests so I'll close this in favour of that one. That should hopefully be merged shortly.

PR #212, opened 2015-08-19
Allow snapshot to be specified in block mappings

cpick:
This change makes it possible to attach volumes based on snapshots to new ec2
instances.


DavidS:
Hi Chris, this change looks good, but would clutter the git history in its current state. Please squash them together and repush here.

DavidS:
It would also be good to add a short note to the README describing the new property.

cpick:
Documented and squashed.  Should be good to go.

DavidS:
Thank you for your contribution!

PR #217, opened 2015-08-26
Filter subnet search based on AZ

withnale:
At our organisation we have our subnets with a common name across AZs (not interested in the pros and cons of this - it wasn't my decision!). However, it means that often puppet finds the wrong subnet.

Considering it's necessary to specify the AZ on ec2_instance creation, it's okay to filter the subnet search based on the AZ.


withnale:
Urgh.

PR #218, opened 2015-08-28
Add support for region and proxy facts

aharden:
When using this module during a Puppet agent run, the environment variables AWS_REGION and PUPPET_AWS_PROXY are not available to Puppet.  This adds support for facts named aws_region and puppet_aws_proxy that can be used for the same purpose.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,9 @@ region using an environment variable.
 export AWS_REGION=eu-west-1
 ~~~
 
+Alternatively, an `aws_region` fact can be used when environment variables 
+aren't accessible (such as during Puppet agent runs).
+
 #### A note on proxies
 
 By default the module accesses the AWS API directly, but if you're in an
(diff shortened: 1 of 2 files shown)


DavidS:
Hi, please squash this into a single commit to keep the git history clean!

garethr:
I'm not sure facts are the right way to go for configuring this information. They provide a convenient methods but I'm not sure they really represent facts about the host.

We have some WIP to support the configuration file for agents for credentials. This could be expanded to support the related config files for specifying proxy and region details too: https://github.com/puppetlabs/puppetlabs-aws/pull/177

aharden:
I need a different way to configure an arbiter.  Environment variables alone don't support my use case.  I commented on an issue regarding this several weeks ago and didn't get a response, so I figured a PR was a way to get attention. ;)

I'll hold this change to my local fork until a better method is available.  @garethr, is supporting proxy via config file a priority for the next release?

Thanks.

PR #245, opened 2015-10-21
Modifying Route 53 Provider

DoctahPopp871:
Resolving a bug in the route53 provider that was preventing the
retrieval of more than 100 resource_record_sets in any given zone.
The Change enumerates the aws response object, allowing the client to page
through the responses.  I have been able to test this against a route53 zone with over 900 records. Provider now returns all values.


garethr:
@DoctahPopp871 thanks for submitting, much appreciated.

PR #246, opened 2015-10-26
(FM-2802) Introduce global config and credentials files in Puppet con…

DavidS:
…fdir

This is to work around the Puppet agent nuking environment variables,
meaning the default location (based on HOME) doesn't work correctly.
This also adds support for storing the other settings in global files.


DavidS:
This is an updated version of #177 containing the small fix mentioned there. @garethr has earmarked the README changes for release prep.

PR #250, opened 2015-11-03
Introduce new feature: Manage SQS queues

jae2:
Introducing ability to create, update and destroy SQS queues based on the name of the queue. The only supported queue attributes are:
 - delay_seconds
 - maximum_message_size
 - message_retention_period

Previously there was no support for managing SQS queues.

Unit tests included for both type/provider and also e2e acceptance test is included.

I'm not much of a Ruby developer and haven't written a custom type before so it would be great to get some feedback!

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,7 @@ You can use the aws module to audit AWS resources, launch autoscaling groups in 
 * `route53_srv_record`: Sets up a Route53 SRV record.
 * `route53_txt_record`: Sets up a Route53 TXT record.
 * `route53_zone`: Sets up a Route53 DNS zone.
+* `sqs_queue`: Sets up an SQS queue.
 
 ###Parameters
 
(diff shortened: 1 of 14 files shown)


jae2:
Looks like I messed things up by running my provider spec in us-west-1. Will fix that up..

garethr:
@jae2 just to say this looks fantastic, thanks so much, especially for going the extra mile with the acceptance tests and docs. I don't have time quite yet to review properly but I'll make some time as soon as I can - I'd love to merge this in and on first glance it looks great. Hopefully back to you soon.

garethr:
@jae2 thanks again for this. I've had a run through the code and checked out the acceptance tests (thanks again for adding those.) Everything looks great, much appreciated. We'll be releasing a 1.3.0 version of the module and this should be included in that.

PR #251, opened 2015-11-09
(FM-3772) Resolve issue when using running for ensure on ec2_instance

garethr:
The logic around the use of running for ensure was incorrect. Although
instances would correctly report they were running, if you used
'running' rather than 'present' in your manifests it would mistakenly
issue change notifications.

This commit fixes the logic and adds (previously failing) tests to
verify the change.

### DIFF
--- a/lib/puppet/type/ec2_instance.rb
+++ b/lib/puppet/type/ec2_instance.rb
@@ -22,9 +22,11 @@ Puppet::Type.newtype(:ec2_instance) do
       current == desired ? current : "changed #{current} to #{desired}"
     end
     def insync?(is)
-      is = :present if is == :running
-      is = :stopped if is == :stopping
-      is.to_s == should.to_s
+      is.to_s == should.to_s or
+        (is.to_s == 'running' and should.to_s == 'present') or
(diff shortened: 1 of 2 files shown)


lib/puppet/type/ec2_instance.rb
```diff
@@ -22,9 +22,10 @@ def change_to_s(current, desired)
       current == desired ? current : "changed #{current} to #{desired}"
     end
     def insync?(is)
-      is = :present if is == :running
-      is = :stopped if is == :stopping
-      is.to_s == should.to_s
+      is.to_s == should.to_s or
+        (is.to_s == 'running' and should.to_s == 'present' ) or
+        (is.to_s == 'stopped' and should.to_s == 'present' ) or
+        (is.to_s == 'stopping' and should.to_s == 'stopped' )
```
DavidS: Assuming this is transitive, the last two lines would imply that `is.to_s == 'stopping' and should.to_s == 'present'` also needs to be added.

garethr: Agreed. Good catch. Fixed and pushed.

PR #255, opened 2015-11-19
(MAINT) Fix issue where an exception triggers another exception

garethr:
This is a copy-paste bug. Route53 is a global service (it's DNS), unlike
all the other services which are scoped to region. This is rarely seen,
but a user managing a large number of records hit it a few times. This
should mean when the exception occurs a useful error message is produced
rather than getting Error: Failed to apply catalog: undefined local
variable or method `region' for.


DoctahPopp871:
Implemented this change in our environment, and are receiving much more useful error messages from Route53.  Can confirm the actual error is exceeding the API rate limit

Issue #264, opened 2015-12-16
ec2_instance reports incorrect region metaparameter when using config file


aharden:
When setting `region = us-east-1` in  `$confdir/puppetlabs_aws_configuration.ini` as documented in https://github.com/puppetlabs/puppetlabs-aws/pull/263 and running `puppet resource ec2_instance`, I get a list of our current EC2 instances in us-east-1, but each of them have a `region` parameter of `eu-west-1`.  When running the `puppet resource` command with both the REGION environment variable and configuration file region set to `us-east-1`, the correct `us-east-1` metaparameter is shown for the listed instances.

daveseff:
Submitted PR #275 that fixes this issue.

aharden:
@daveseff Thanks!

DavidS:
Merged. Thanks everyone!

PR #275, opened 2016-02-08
Fixed issue #264 where region misreports as eu-west-1 when AWS_REGION…

daveseff:
… is not set


DavidS:
Shouldn't this be the other way around, the ENV var taking precedence over the global configuration?

daveseff:
This was a scenario where there is no local ENV var specified, but was defined in the global config and therefore uses a hardcoded default. But I agree with you where ENV should override so I've updated my commit. Cheers,

Issue #276, opened 2016-02-09
ec2_instance source/dest check


mlosapio:
ec2_instance type should support source_dest_check and should not be a 1 time creation parameter.

mlosapio:
So this turned out to be a huge pain since AWS doesn't let you launch an instance with Source/Dest check specified. It's only available after launch - and it's sort of weak to require puppet run twice to have things configured correctly.

ahenroid:
@mlosapio, after closing #277, have you concluded that this is unworkable? Should we close this issue for now or leave it open if anyone wants to pick it up? Thanks!

mlosapio:
I just dont have the time to pick it back up - AFAIK its still an open issue.

ahenroid:
Thanks, @mlosapio. I'll leave this on the books.

PR #277, opened 2016-02-09
Added source/dest check parameter for the ec2_instance type

mlosapio:
https://github.com/puppetlabs/puppetlabs-aws/issues/276

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -88,6 +88,7 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
       security_groups: instance.security_groups.collect(&:group_name),
       subnet: subnet_name,
       ebs_optimized: instance.ebs_optimized,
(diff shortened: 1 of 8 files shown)


mlosapio:
I can't figure out vcr and what's wrong here... if someone can help me out I'd appreciate it.

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -293,11 +294,28 @@ def create
         )
       end
 
+      ec2.wait_until(:instance_running, instance_ids: instance_ids)
```
hanikesn: Can't we only wait when we actually want to change the value or do we have to wait in all cases?

zleswomp:
@mlosapio I can help with VCR, but we need to start with a rebase on this one.  Also, more information in the commit would help me understand what the goal here is.  Do you have interest in coming back to this work?

mlosapio:
We can abandon, I dont have the cycles at the moment to come back to this...

zleswomp:
Thanks for coming back with that.  Apologies that this sat for so long @mlosapio.

PR #296, opened 2016-04-11
(CLOUD-498) add minimal tags support to ec2_autoscalinggroup

DavidS:
ec2_autoscalinggroup has a different API for tags compared to other
resources. Presumably this is because of the propagate_at_launch
complexity.

The changes are covered by acceptance tests, while the VCR-based specs
were beyond repair. When deleting the cassettes, one of the tests didn't
even have a cassette recorded. I don't know what that was testing.


gregohardy:
+1 on the change. The code looks good.

Issue #227, opened 2015-09-23
Error: Could not run: Puppet detected a problem with the information returned from AWS


anishnath:
[root@puppet XYZ]# puppet resource cloudwatch_alarm -y
Error: Could not run: Puppet detected a problem with the information returned from AWS
when looking up cloudwatch_alarm in us-east-1. The specific error was:

Malformed policy ARN. Given: arn:aws:sns:us-east-1:754261010230:testuser

Rather than report on cloudwatch_alarm resources in an inconsistent state we have exited.
This could be because some other process is modifying AWS at the same time.

All other resources are working properly expect the cloud_watch

benglersc:
I'm getting the same error message. It looks like the cloudwatch_alarm resource currently does not support the Notification Action.

anishnath:
AWS module is great we need support for the following tooSSL Configuration on the loadbalancer Changing the Instance Type Enable Disable Source/Dst CheckCloud Watch  
  


     On Thursday, 5 November 2015 1:10 AM, Brian Engler <notifications@github.com> wrote:
   

 I'm getting the same error message. It looks like the cloudwatch_alarm resource currently does not support the Notification Action. —

agarstang:
Reproduced this on Puppet 4.3.2. Using IAM role to assign permissions to Puppetmaster, no combination of policies allowed me to list 'cloudwatch_alarm' resources.

```
~# puppet resource cloudwatch_alarm --debug --verbose
Debug: Runtime environment: puppet_version=4.3.2, ruby_version=2.1.8, run_mode=user, default_encoding=UTF-8
Debug: Evicting cache entry for environment 'production'
Debug: Caching environment 'production' (ttl = 0 sec)
Error: Could not run: Puppet detected a problem with the information returned from AWS
when looking up cloudwatch_alarm in eu-west-1. The specific error was:

Malformed policy ARN. Given: arn:aws:sns:eu-west-1:018637733392:dynamodb

Rather than report on cloudwatch_alarm resources in an inconsistent state we have exited.
This could be because some other process is modifying AWS at the same time.
```

agarstang:
Reproduced this on Puppet 4.3.2. Using IAM Role to assign permissions to Puppetmaster, no combination of polices allowed me to list 'cloudwatch_alarm' resources.

```
~# puppet resource cloudwatch_alarm --debug --verbose
[...]

PR #300, opened 2016-05-05
provider/cloudwatch_alarm: do not convert arns into autoscaling groups

fsouza:
The way I detect that it's a arn is by checking whether the string
starts with arn:aws. There are probably better ways of doing that. We
could be more specific and support only certain kinds of arns, for
example.

I failed to change/prepare a test for this change. I'd love to include a
test in this change, if someone could point me to the right direction.

This is related to #227.

### DIFF
--- a/lib/puppet/provider/cloudwatch_alarm/v2.rb
+++ b/lib/puppet/provider/cloudwatch_alarm/v2.rb
@@ -33,10 +33,15 @@ Puppet::Type.type(:cloudwatch_alarm).provide(:v2, :parent => PuppetX::Puppetlabs
   read_only(:region, :alarm_actions)
 
   def self.alarm_to_hash(region, alarm)
-    response = autoscaling_client(region).describe_policies(
-      policy_names: alarm.alarm_actions
-    )
-    actions = response.scaling_policies.collect(&:policy_name)
+    actions = alarm.alarm_actions
+
+    unless actions.first.start_with?('arn:aws')
+      response = autoscaling_client(region).describe_policies(
+        policy_names: actions
+      )
+      actions = response.scaling_policies.collect(&:policy_name)
+    end
+
     {
       name: alarm.alarm_name,
       metric: alarm.metric_name,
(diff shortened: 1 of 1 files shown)


DavidS:
I think I understand what's going on here: the API and the manifests use different concepts to refer to alarms, and the original code seems to be confused about that. In the same vein I'm not convinced that the conditionals are required here. e.g. `alarm_to_hash` should just know what gets passed in (values from the API). The original code is also passing all tests (in spec/acceptance/autoscaling_spec.rb).

I'll be offline for the next two weeks, so I can't currently give this as much attention as it should have. If you could have a look and try to build a acceptance test that demonstrates the cases inwhen the original error happens? Or, even just modify the current ASG tests to hit that issue, would already be a great step forward, so I can pick it up when I return.

fsouza:
@DavidS alarm_to_hash makes a call to describe autoscaling policies , and AWS fails saying that the provided value is not a valid asg identifier. The test suite mocks AWS API, it isn't capable of ensuring that the library behaves correctly with the actual API. I wasn't able to add a new test case with the error response form the AWS API, but if you can point me to the right direction, I can have a look.

As described in https://github.com/puppetlabs/puppetlabs-aws/issues/227#issuecomment-205213335, the error messages says "Malformed policy ARN.".

lib/puppet/provider/cloudwatch_alarm/v2.rb
```diff
@@ -33,10 +33,15 @@ def self.prefetch(resources)
   read_only(:region, :alarm_actions)
 
   def self.alarm_to_hash(region, alarm)
-    response = autoscaling_client(region).describe_policies(
-      policy_names: alarm.alarm_actions
-    )
-    actions = response.scaling_policies.collect(&:policy_name)
+    actions = alarm.alarm_actions
+
+    unless actions.first.start_with?('arn:aws')
```
hunner: Is it reasonable to only inspect the first action, or should all actions be inspected for `arn:aws` when collecting the `policy_name`?

zleswomp: @fsouza I also have this question.

fsouza: Hey, yes, we should probably inspect all actions. What should the code do in this case? Collect all the actions that are not arn as pass as parameter  to describe_policies? Then merge "actions" back?

I can work on that if we agree with this solution. Sorry for taking almost one month to respond :(

PR #307, opened 2016-05-19
Add initial iam_user resource

zleswomp:
This work builds on the examples provided by other types and replicates
the work into a new type for 'iam_user'.  Without this change, it is not
possible to manage IAM users in AWS using puppet.

### DIFF
--- a/fixtures/vcr_cassettes/create-user.yml
+++ b/fixtures/vcr_cassettes/create-user.yml
@@ -14,13 +14,13 @@ http_interactions:
       User-Agent:
       - aws-sdk-ruby2/2.0.5 ruby/2.2.3 x86_64-darwin15
       X-Amz-Date:
-      - 20160520T211454Z
+      - 20160523T185231Z
       X-Amz-Content-Sha256:
       - b6359072c78d70ebee1e81adcbab4f01bf2c23245fa365ef83fe8f1f955085e2
       Authorization:
-      - AWS4-HMAC-SHA256 Credential=redacted/20160520/us-east-1/iam/aws4_request,
+      - AWS4-HMAC-SHA256 Credential=redacterd/20160523/us-east-1/iam/aws4_request,
         SignedHeaders=content-type;host;user-agent;x-amz-content-sha256;x-amz-date,
-        Signature=270839ae34b9e857252ec574d90da7a1396864421914927c612f6bc938dbcbb3
+        Signature=df19351c465e533b6168da0530af88b1e87f464e8fe7324a0cb97abb13bac5c8
       Content-Length:
       - '35'
       Accept:
@@ -31,13 +31,13 @@ http_interactions:
       message: OK
     headers:
       X-Amzn-Requestid:
-      - e61b2219-1ecf-11e6-b905-2f308342ffb7
+      - 812d8fd4-2117-11e6-a813-d3092ea6021a
       Content-Type:
       - text/xml
       Content-Length:
-      - '1121'
+      - '617'
       Date:
-      - Fri, 20 May 2016 21:14:54 GMT
+      - Mon, 23 May 2016 18:52:31 GMT
     body:
       encoding: UTF-8
       string: |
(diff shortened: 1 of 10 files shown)


zleswomp:
@DavidS Hi.  How do I go about recording a VCR cassette to test the provider here?

zleswomp:
@DavidS Never mind, I had to work out the diplomacy here to create a new account under my employers name for this work.  I've been able to capture the requests in VCR.

Review here would be appreciated.

zleswomp:
Strange.  The tests pass locally.  I'll take another pass.

zleswomp:
I've rebased locally and found that the version of the sdk update that has happened since this was opened has introduced a change in behavior.  I'm working it out locally.

zleswomp:
This error may be a bug introduced in the sdk upstream.  Running an older version of the sdk does not produce this error.

DavidS:
@zleswomp thanks for your tenacity here. The VCR based tests are not the most robust setup, and we're currently considering ditching them completely. Your issues with them only supports that point of view.

The next question will be whether the code is covered by acceptance tests (see `spec/acceptance` folder).

I also note that the type currently only covers the bare minimum of properties of the IAM user. Do you not need any of the other properties (like policies, groups, etc.)?

zleswomp:
@DavidS Thank you for looking here.  I don't currently need the other properties on the user account. It looks like the API docs for `create_user` would only allow the 'path' as well, which I don't require at this time.  My goal here was a) to get user account provisioning working with puppet and b) ensure that this code could indeed make it into the module.  I can confirm that this is indeed working, but I'm not sure what to do about the tests.

If we can sort out the tests to arrive at a merge state for this code, I can continue down the path of polices etc, but this work on its own provides (me, and therefore, my employer) value and I didn't want to spend the effort at the risk of the changes being rejected, etc.  Also wanted to get familiar with testing, which I have done now.

I'm open to suggestions here on how to get this work accepted.  I've not looked at acceptance testing.  Some brief docs would be helpful in this are on how I would go about contributing quality code to this repo.  I have a stack of work coming that this module could support, but doesn't currently.

DavidS:
@zleswomp That's fine, don't worry too much about it.The functionality looks fine for a first shot and can be merged, once the test failure is addressed. That looks like the sdk is not happy with data your putting in.

There are currently no good acceptance test docs I can point you to, but you can have a look at e.g. https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/acceptance/autoscaling_spec.rb which has good test coverage of the ASG type. The core goal is to demonstrate that the type is really changing AWS as intended, and that puppet resource works

zleswomp:
@DavidS I'll have a read over the link for acceptance testing and see if I can get something working.  In that case, should I remove the broken unit tests?  Its worth noting that I was able to get the tests to pass locally before the update to the SDK gem made it in and this code does actually work in its current form, so I'd like to do whatever testing updates are needed to get this merged.  I'm expecting to follow up in the next couple week with some iam policy work as well, though I'm not sure what that will look like yet, so figuring out testing before I begin that would would be ideal for me.

zleswomp:
@DavidS I've taken a stab at writing acceptance tests.  I don't actually know how to run them, but they might work... :)  When I ran teh `bundle exec rake acceptance` things just blew up all over my screen.  I checked the account I'm using for this work and nothing seemed to get created, so I wasn't sure if the environment variables for AWS were being read from my shell's environment.  In any case, have a review if you would please and let me know how you would like to proceed and how I can help move this along.

DavidS:
Hi @zleswomp , apologies for my slow response, I've been on holidays in between. I'll have a look once I work through the rest of my inbox. Until then you might want to take a look at the unit test failure (https://travis-ci.org/puppetlabs/puppetlabs-aws/jobs/133255978#L857) where apparently the code seems to be hitting some validation in the aws sdk?

zleswomp:
Hi @DavidS, That error looks to me to be caused by VCR itself, and only after the recent update of the `aws-sdk-core` gem.  The code itself I believe is correct, and did pass tests until I rebased to bring in the sdk gem change as mentioned.  Also, running `bundle exec puppet apply` with a test manifest does indeed do the correct thing, so the code works here, this is just an issue with testing.  It doesn't look like VCR has had a newer version cut since the build failure you've linked, so I'm not sure where to proceed.  

If there is not a reasonable way to get the tests to pass here, I'm willing to remove all tests from this PR to help move this forward.  What do you think?

DavidS:
Those VCR tests are a pain. Feel free to remove them (and the associated
VCR files) to get this to pass. We can still rely on the acceptance tests.

zleswomp:
I've built more types and providers in https://github.com/puppetlabs/puppetlabs-aws/pull/316

If that looks good to you, we can close this one out in favor of merge there.

PR #313, opened 2016-05-28
Launch configuration/block device mappings

benpillet:
I have the implementation in there.  But, I'm not sure how to do the VCR unit testing.  I added the beginnings of hte test in spec/unit/provider/ec2_launchconfig/v2_spec.rb but I'm open to guidance on how to update the cassettes.


DavidS:
Thanks for your contribution!

PR #317, opened 2016-07-11
Make SSD-based EBS volumes the default.

msilvey:
Without this change all EBS volumes provisioned by the the puppetlabs aws code will default to magnetic media for backing.  This change will make the use of SSD-based data storage default.

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -223,7 +223,7 @@ Found #{matching_groups.length}:
           volume_size: device['volume_size'],
           snapshot_id: device['snapshot_id'],
           delete_on_termination: device['delete_on_termination'] || true,
-          volume_type: device['volume_type'] || 'standard',
+          volume_type: device['volume_type'] || 'gp2',
           iops: device['iops'],
           encrypted: device['encrypted'] ? true : nil
         },


hunner:
Eh, this would be a semver-breaking change so I don't think we can incorporate this yet. What do you think?

msilvey:
I don't understand what you mean.  Explain please?

msilvey:
Where this gets a little messy, the AWS marketing docs indicate that gp2 is the default and indicates that 'standard' type magnetic media is a previous generation.
ex: https://aws.amazon.com/ebs/details/

but the API documentations indicates that standard is the default.
ex: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateVolume.html

From my personal viewpoint, I would say that I haven't desired to use HDDs outside certain data warehouse applications for quite a while.  I can't say how the community would view that.

From a cost (in USD) POV: 
The 'gp2' type defaults to 0.10 cents / GB / month.
The 'sc1' type ('standard' replacement) defaults to 0.025 cents / GB / month.
The 'standard' type defaults to 0.05 cents / GB / month.

Any user of the 'standard' default is effectively handing over a 2.5 cents extra per GB for the same performance to be had in a comparable type.

refs:
https://aws.amazon.com/ebs/previous-generation/
https://aws.amazon.com/ebs/details/

hunner:
Thanks for putting it into perspective! If the community has adopted gp2 as the go-to medium, then that would make doing a major bump for this change reasonable.

PR #281, opened 2016-02-19
Support for managing EC2 Volumes

daveseff:
I am adding support for managing EC2 Block device volumes. With this you can create and attach volumes to already running instances as well as detach and destroy. 

Volumes can be created from scratch or built from a snapshot. 

TODO: 
- Add the ability to reference ec2_instances by name. I found that doing ec2_instance name lookups is really slow so for performance I left it out. If someone knows a better way, please let me know.

PR #320, opened 2016-07-15
Initial Ec2 Volume management

zleswomp:
Without this work, managing volumes independent of Ec2 instances is not
possible using this module.  This is the preliminary work to get
Ec2 Volume management operational under Puppet.

Expect that this will be followed up with adjustments to the
ec2_instance type to allow for attachment of volumes that are managed
outside of the ec2_instance type.  Future work will allow volumes to be
carried from one instance to another as instances are decommissioned,
etc.


zleswomp:
Ah, now I (again) see this duplicates https://github.com/puppetlabs/puppetlabs-aws/pull/281

PR #325, opened 2016-07-27
Amend contributing doc to include VCR example

zleswomp:
VCR can be confusing, and without this work, there is not a clear
example on how to contribute tests whilst using VCR and protecting
credentials.  Here we write up an example and explain a test case to
demonstrate how VCR is intended to be used.  Also we include a note
about how to clean up credential to avoid leaking information.


zleswomp:
I've made no code changes.  The build fails on Ruby 1.9.  Might we merge anyway?

PR #328, opened 2016-07-28
Add alias_target property for Route53 types

zleswomp:
The alias_target of a route53 resource record is a stand in value that
allows resources to shift behind the scenes without requiring action on
the part of the user.  Without this change, we are unable to set or
retrieve the alias_target of a route53 record.  Here we simply add a
alias_target property to enumerate the value if its found to exist.

### DIFF
--- a/lib/puppet/provider/route53_record.rb
+++ b/lib/puppet/provider/route53_record.rb
@@ -8,17 +8,25 @@ class Puppet::Provider::Route53Record < PuppetX::Puppetlabs::Aws
       records = []
       zones_response.data.hosted_zones.each do |zone|
         route53_client.list_resource_record_sets(hosted_zone_id: zone.id).each do |records_response|
-        records_response.data.resource_record_sets.each do |record|
-          records << new({
-            name: record.name,
-            ensure: :present,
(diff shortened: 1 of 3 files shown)


hunner:
This also needs an ability to set the alias in create/flush.

zleswomp:
Thank you @hunner.  This took me a little longer than I was expecting, but it all works.  Let me know if the docs aren't clear, or the `fail()` messages don't sound right.  I need a `conflictswith` param or something.  Anyway, maybe this is good now.

lib/puppet/provider/route53_record.rb
```diff
@@ -39,23 +47,48 @@ def exists?
   end
 
   def record_hash(action)
-    values = resource[:values] || @property_hash[:values]
+    values = resource[:values]
```
zleswomp: Someone please tell me if this is bad.  I don't know why we would ever take the value from the discovered resource.

hunner: I think resource[:values] would actually be set from the @property_hash if no value was passed in, so net effect the same.

PR #330, opened 2016-07-28
Bump the SDK version to accommodate Ruby 1.9

zleswomp:
Changes outside of this repo have caused some failures with the
installation of json_pure on Ruby 1.9.  Here just seeing what happens
when we bump to a new version of the SDK.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ def location_for(place, fake_version = nil)
   end
 end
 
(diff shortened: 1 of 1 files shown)


zleswomp:
This is not required for the ruby 1.9 fix, though, its worth noting that the tests pass.  I can update the commit message if this change is desirable.

hunner:
No longer needed

PR #332, opened 2016-07-28
Pin json_pure for ruby 1.9

zleswomp:
Without this change, building on ruby 1.9 complains about the latest
version of json_pure requiring Ruby 2.x.  Here we pin the version of
json_pure in the case where RUBY_VERSION is set to 1.9.3.  This should
allow us to build on both 1.9 and 2.x.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ group :test do
   gem 'vcr'
   gem 'rspec-puppet', :git => 'https://github.com/rodjek/rspec-puppet.git'
   gem 'metadata-json-lint'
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -21,6 +21,7 @@ group :test do
   gem 'vcr'
   gem 'rspec-puppet', :git => 'https://github.com/rodjek/rspec-puppet.git'
   gem 'metadata-json-lint'
+  gem 'json_pure', '1.8.3' if RUBY_VERSION == '1.9.3'
```
hunner: Could you make this `~> 1.0` in case there are future 1.x releases?

PR #324, opened 2016-07-27
Add ECS resources

zleswomp:
This begins the management of AWS Ec2 Container Service resources using Puppet.

PR #334, opened 2016-07-29
Replace read-only failure with warning

zleswomp:
Currently, if a read-only property is called upon to make a change to an
existing resource, a fail() is called to throw an error to the user,
even though the rest of the run continues.  Notifying the user that the
change will not be made, while definitely useful, is not a failure
condition when operating AWS infrastructure.  If a change is made to
either the resource in Puppet, or the live resource in AWS, which
results in a difference in desired and actual state for one of these
properties, this condition is raised.

To fill in the gaps of what this module currently provides and what is
necessary to be done outside of Puppet, I believe a warning is much more
appropriate, as we still notify the user that a change will not be made,
but also allow for clean(ish) catalog runs so reports report clean.

### DIFF
--- a/lib/puppet_x/puppetlabs/aws.rb
+++ b/lib/puppet_x/puppetlabs/aws.rb
@@ -52,7 +52,7 @@ This could be because some other process is modifying AWS at the same time."""
       def self.read_only(*methods)
         methods.each do |method|
           define_method("#{method}=") do |v|
-            fail "#{method} property is read-only once #{resource.type} created."
+            Puppet.warning "#{method} property is read-only once #{resource.type} created."
           end
         end
       end


zleswomp:
@DavidS @hunner What do you guys think about this one?  A merge here would allow me to move my production infrastructure to the ECS branch I've got open for https://github.com/puppetlabs/puppetlabs-aws/pull/324.

DavidS:
@zleswomp The code should fail that and only that resource, because it is out of compliance. I don't think that only outputting a warning would be appropriate there.

zleswomp:
@DavidS If Puppet is incapable of rectifying the situation, then can you also see the argument that it is not Puppet that should be failing here?  In other circumstances (local systems changes, etc) I would completely agree with your assessment.  Though, creating a thing in aws and not being able to modify it is just something that you have to live with in aws, and is a limitation of the API and AWS itself.  Its not really the responsibility of Puppet to fix the issue once the resource for a read_only property has been set.  I can see for auditing etc, but failing the resource for something that Puppet has no actual control over seems a little strange to me.  Also noisy.

daveseff:
I think a by-product of this is now in the puppet enterprise console this "change warnings" are being reported as actual changes.

PR #340, opened 2016-08-05
New provider and types: iam_instance_profile, iam_role

reubenavery:
In order to satisfy our use case of creating IAM roles which can be assigned to EC2 instances, there is the need to introduce the iam_instance_profile resource type, and an associated iam_role resource type.

### DIFF
--- a/lib/puppet/provider/iam_instance_profile/v2.rb
+++ b/lib/puppet/provider/iam_instance_profile/v2.rb
@@ -1,0 +1,53 @@
+require_relative '../../../puppet_x/puppetlabs/aws.rb'
+
+require 'uri'
+
+Puppet::Type.type(:iam_instance_profile).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) do
+  confine feature: :aws
+
+  mk_resource_methods
+
+  def self.instances
+    response = iam_client.list_instance_profiles()
+    response.instance_profiles.collect do |instance_profile|
+      new({
+              name: instance_profile.instance_profile_name,
+              ensure: :present,
+              path: instance_profile.path,
+              arn: instance_profile.arn,
+              roles: instance_profile.roles,
+          })
+    end
+  end
+
+  def self.prefetch(resources)
+    instances.each do |prov|
+      if resource = resources[prov.name]
+        resource.provider = prov
+      end
+    end
+  end
+
+  def exists?
+    Puppet.info("Checking if IAM instance profile #{name} exists")
+    @property_hash[:ensure] == :present
+  end
+
+  def create
+    Puppet.info("Creating IAM instance profile #{name}")
+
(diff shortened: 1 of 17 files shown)


lib/puppet/provider/iam_instance_profile/v2.rb
```diff
@@ -0,0 +1,92 @@
...
+  def self.prefetch(resources)
+    instances.each do |prov|
+      if resource = resources[prov.name]
+        resource.provider = prov
+      end
+    end
+  end
+
+  def exists?
+    Puppet.info("Checking if IAM instance profile #{name} exists")
```
zleswomp: I've replaced all the rest of these to be `debug`.  Would you mind updating these messages to use `debug` also please?

reubenavery:
yeah, I'm relying on IntelliJ's code formatting, and these huge indents for continuation blocks (is that what they're called?) seems to be consistent with the existing code base so I left it.

zleswomp:
Its fine enough for me.  Thanks for putting in the effort @reubenavery.

reubenavery:
OK, bumped down most of the Puppet.info to Puppet.debugs.

zleswomp:
Excellent, thank you.  I'm still +1 on this change.

zleswomp:
Maybe it makes sense to uncomment here for this change also? https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/iam_policy_attachment.rb#L12

lib/puppet/provider/iam_instance_profile/v2.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def self.instances
+    response = iam_client.list_instance_profiles()
+    response.instance_profiles.collect do |instance_profile|
+      new({
+              name: instance_profile.instance_profile_name,
+              ensure: :present,
+              path: instance_profile.path,
+              arn: instance_profile.arn,
+              roles: instance_profile.roles,
```
hunner: On puppet 3 the type with `:array_matching => :all` will return a single non-array value whether it is passed a single value or an array of a single value. On puppet 4 the type will always return an array of single or multiple values.

The problem is that if your provider `self.instances` returns roles as an array of a single item, puppet 3 will try to turn `['value']` into `'value'` and cause changes every run. You can document this property as only working on puppet 4. Making it work on 3 is just icky.

hunner:
Our CI currently does not have the ability to manage IAM credentials, so merging this will cause our acceptance tests to fail. I'll raise a ticket with our QA team to work through this before merging.

zleswomp:
If the acceptance tests are removed from this PR, is it acceptable to merge?  I'd like to use the iam_role type from this work.

hunner:
Maybe just comment out the tests in case we ever can test them?

PR #345, opened 2016-08-10
Add health check management on ELBs

zleswomp:
Without this change, there is no ability to manage the configuration of
the health_check property on an ELB.  Here we add support for
configuring the healthcheck properties of an ELB.


zleswomp:
Thank you @eputnam!

PR #352, opened 2016-08-23
Fix IAM policy enumeration and centralize

zleswomp:
Without this change, the IAM polices are not completely enumerated
during prefetching due a failure to inspect the #is_truncated method on
the results returned.  This work centralizes the logic to perform the
enumeration into an iam_policy helper module and calls this enumeration
from both the iam_policy provider as well as the iam_policy_attachment
provider.  This should ensure that both providers contain complete
results when enumeration of the existing policies is called upon.


zleswomp:
The failure here looks like something outside my code.  Maybe a retry on travis?

PR #356, opened 2016-09-01
Add param to skip image comparison on ECS container

zleswomp:
Without this change, the behavior of puppet is to take all container
properties into consideration during comparison what is with what should
be.  Normally this makes a great deal of sense, but becomes a problem in
environments where an external tool (as part of CI) modifies the image
of a container to reflect the current desired build.  This results in
puppet overwriting the image due to correctly taking the responsibility
of managing the entire container definition.

This change adds a new parameter to the ecs_task_definition type called
'replace_image', which when set to false, will copy the existing image
value from the existing container into the container that puppet reports
"should" be.  This has the effect of ignoring the image that has been
set in Puppet, while still allowing the comparison for all other aspects
of a container to retain existing comparison functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -987,6 +987,14 @@ above.
 ##### `volumes`
 An array of hashes to handle for the task.
 
+##### `replace_image`
+A boolean to turn off the replacement of container images.  This enables Puppet
+to create, but not modify the image of a container once created.
+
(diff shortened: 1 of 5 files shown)


zleswomp:
Thank you @hunner

PR #361, opened 2016-09-08
Handle truncated results for iam_user enumeration

zleswomp:
Without this change, truncated results from the AWS API are not handled
correctly, leaving some user resources undiscovered.  This results in an
incomplete picture of what already exists.  Here we extend the iam_user
provider to cleanly handle truncated results and continue enumerating.

### DIFF
--- a/lib/puppet/provider/iam_user/v2.rb
+++ b/lib/puppet/provider/iam_user/v2.rb
@@ -5,9 +5,27 @@ Puppet::Type.type(:iam_user).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) d
 
   mk_resource_methods
 
+  def self.get_users
+    user_results = iam_client.list_users()
+    users = user_results.users
+
+    truncated = user_results.is_truncated
+    marker = user_results.marker
+
+    while truncated and marker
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/iam_user/v2.rb
```diff
@@ -5,9 +5,29 @@
...
+    user_results = iam_client.list_users()
+    users = user_results.users
+
+    truncated = user_results.is_truncated
+    marker = user_results.marker
+
+    while truncated and marker
+      Puppet.debug('iam_user results truncated, proceeding with discovery')
+      response = iam_client.list_users({marker: marker})
+      response.users.each {|u|
```
bastelfreak: ruby styleguide recommends using a oneliner here AFAIK:
```ruby
response.users.each {|u| users << u }
```

bastelfreak: multiline do/end is also possible, but should only be used if you do multiple statements in it, like this:

```ruby
response.users.each do user
  users << user
  puts user
end
```

zleswomp: Updated, thank you @bastelfreak.

zleswomp:
Thanks @hunner

Issue #209, opened 2015-08-13
Allow rds_instance creation to accept a snapshot parameter


dfuentes77:
so that it can create an RDS instance restored from another db's snapshot like what "aws rds restore-db-instance-from-db-snapshot" would do via command-line tool

ahenroid:
@dfuentes77 @daveseff, I believe this was resolved by #363. Can I close this issue?

dfuentes77:
👍

PR #363, opened 2016-09-13
Rds snapshot restore

daveseff:
Adds support for issue #209. Adding the parameter creates the RDS instance using a DB snapshot.

  restore_snapshot => 'some-snapshot-name'

### DIFF
--- a/lib/puppet/provider/rds_instance/v2.rb
+++ b/lib/puppet/provider/rds_instance/v2.rb
@@ -113,8 +113,15 @@ Puppet::Type.type(:rds_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
       availability_zone: resource[:availability_zone],
     }
 
-    rds_client(resource[:region]).create_db_instance(config)
-
+    if resource[:restore_snapshot]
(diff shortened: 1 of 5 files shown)


zleswomp:
+1, though there seems to be more in this PR than is in the is in the commit messages.

daveseff:
Sorry, I think that was from a bad rebase. I have corrected the PR

zleswomp:
One question for my understanding, but still +1 on this change.

lib/puppet/provider/rds_instance/v2.rb
```diff
@@ -113,8 +113,15 @@ def create
       availability_zone: resource[:availability_zone],
     }
 
-    rds_client(resource[:region]).create_db_instance(config)
-
+    if resource[:restore_snapshot]
+      Puppet.info("Restoring DB instance #{name} from snapshot #{resource[:restore_snapshot]}")
+      [:engine_version, :backup_retention_period].each { |k| config.delete(k) }
```
daveseff: That is correct. The RDS engine version must be the same as the snapshot was created with.

daveseff:
Can this get merged please?

zleswomp:
I'm still +1 on this change.

PR #368, opened 2016-09-23
Add support for modifying key attributes of RDS

zleswomp:
Without this change, the RDS instances that are created with the
rds_instance type are not modifyable.  Puppet detects the changes
correctly, but does nothing to update the real world state.  Here we
take a few important parameters that one might want to modify on the
running rds_instance resources, and flush the changes from the setter
methods.

### DIFF
--- a/lib/puppet/provider/rds_instance/v2.rb
+++ b/lib/puppet/provider/rds_instance/v2.rb
@@ -5,6 +5,11 @@ Puppet::Type.type(:rds_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
 
   mk_resource_methods
 
+  def initialize(value={})
+    super(value)
+    @property_flush = {}
+  end
+
   def self.instances
     regions.collect do |region|
       instances = []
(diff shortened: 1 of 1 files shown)


daveseff:
@zleswomp  I am now getting

Could not evaluate: undefined method `keys' for nil:NilClass

in my puppet runs

zleswomp:
@daveseff May I get more output from you?  If you add `--trace --debug`, you should get a stack trace.  There is only one `keys` call that I see in this code, and above it, we already check a key, so I'm confused.  Could post a gist or something.

zleswomp:
@daveseff I misread my change, confusing the `@property_hash` vs `@propery_flush`.  Could you test https://github.com/puppetlabs/puppetlabs-aws/pull/384 please @daveseff?

Also, can someone explain what I'm clearly not seeing.  In the `initialize` method, we set `@property_flush` to an empty hash, so `.keys` should be fine to call on the object without checking if its nil.

daveseff:
@zleswomp Much better, thanks.

zleswomp:
@daveseff Thanks for letting me know.  Sorry for the trouble.

PR #369, opened 2016-09-30
Add policy management support to s3_bucket type

zleswomp:
Without this change, the s3_bucket type does not have the facility to
manage the policy on a bucket.  Here we add a new property to the type
to allow management of bucket policies with some JSON validation and
nicer representation when using 'puppet resource'.

### DIFF
--- a/lib/puppet/provider/s3_bucket/v2.rb
+++ b/lib/puppet/provider/s3_bucket/v2.rb
@@ -7,17 +7,31 @@ Puppet::Type.type(:s3_bucket).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) 
 
   def self.instances
     response = s3_client.list_buckets()
-    response.buckets.collect do |bucket|
-      new({
+    bucket_list = response.buckets.collect do |s3_bucket|
+
+      data = {
         ensure: :present,
-        name: bucket.name,
-        creation_date: bucket.creation_date,
-      })
+        name: s3_bucket.name,
+        creation_date: s3_bucket.creation_date,
+      }
+
+      begin
+        results = s3_client.get_bucket_policy({bucket: s3_bucket.name})
+        policy_data = JSON.parse(URI.unescape(results.policy.string))
+        policy_document = JSON.pretty_generate(policy_data)
+        data[:policy] = policy_document
+      rescue Exception => e
+        Puppet.debug("An error occurred reading the policy on S3 bucket #{s3_bucket.name}: " + e.message)
+      end
+
+      new(data)
     end
+    bucket_list.reject {|b| b.nil? }
   end
 
   def self.prefetch(resources)
     instances.each do |prov|
+      next if prov.nil?
       if resource = resources[prov.name]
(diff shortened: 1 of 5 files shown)


lib/puppet/provider/s3_bucket/v2.rb
```diff
@@ -39,4 +53,13 @@ def destroy
     s3_client.delete_bucket({bucket: name})
   end
 
+  def policy=(value)
+    Puppet.info('Replacing bucket policy')
+    Puppet.info(value.class)
```
hunner: These lines should probably be combined into one (or perhaps remove the second?) and be `debug` not `info`

lib/puppet/provider/s3_bucket/v2.rb
```diff
@@ -7,17 +7,31 @@
...
+        policy_data = JSON.parse(URI.unescape(results.policy.string))
+        policy_document = JSON.pretty_generate(policy_data)
+        data[:policy] = policy_document
+      rescue Exception => e
+        Puppet.debug("An error occurred reading the policy on S3 bucket #{s3_bucket.name}: " + e.message)
+      end
+
+      new(data)
     end
+    bucket_list.reject{|b| b.class == nil.class }
```
hunner: `bucket_list.reject { |b| b.nil? }` eh?

lib/puppet/provider/s3_bucket/v2.rb
```diff
@@ -7,17 +7,31 @@
...
   end
 
   def self.prefetch(resources)
     instances.each do |prov|
+      next if prov.is_a? NilClass
```
hunner: `next if prov.nil?` eh?

lib/puppet/provider/s3_bucket/v2.rb
```diff
@@ -7,17 +7,31 @@
...
+        creation_date: s3_bucket.creation_date,
+      }
+
+      begin
+        results = s3_client.get_bucket_policy({bucket: s3_bucket.name})
+        policy_data = JSON.parse(URI.unescape(results.policy.string))
+        policy_document = JSON.pretty_generate(policy_data)
+        data[:policy] = policy_document
+      rescue Exception => e
+        Puppet.debug("An error occurred reading the policy on S3 bucket #{s3_bucket.name}: " + e.message)
```
hunner: Under what conditions would this error happen?

lib/puppet/provider/s3_bucket/v2.rb
zleswomp: Replaced.

lib/puppet/provider/s3_bucket/v2.rb
zleswomp: Replaced.

lib/puppet/provider/s3_bucket/v2.rb
zleswomp: Removed the second, converted to debug.

PR #370, opened 2016-10-07
Delete IAM login profile during user delete

zleswomp:
Without this change, IAM user deletion fails if the account has MFA
tokens, passwords set, or other access keys assigned to the account.

Here we remove the login profile, access keys, and associated MFA tokens
for the user account.  This allows successful removal of the account.


zleswomp:
Hold on.  This is incomplete.  I will update after I've completely removed a user from my account.

zleswomp:
I've updated the commit and the message to accurately reflect the change.  This is the work that was required to remove a user from my environment.

PR #371, opened 2016-10-12
Improve ELB enumeration performance and stuff

zleswomp:
Without this change, when we need to generate a Hash of all the
information about a particular ELB during the resolution of instances,
and in the process, convert any security group, subnet, and instances
that the ELB is attached to from ID strings into resource names,
subsequent calls are made to the API for each of these enumerations.
This presents a problem due to the number of calls made to the API,
which is three calls to the Ec2 API per ELB.  This is inefficient,
because we have an autorequire that ensures the instances calls on the
relevant puppet resources are already enumerated, meaning that much of
the data we require in the ELB hash generation is already present in
memory, but we make the API calls anyway.  This is only true for
resources under Puppet management, as resources not under puppet
management are thrown away during the instances call for a given
provider.  In environments with lots of ELBs, this can mean that the API
limit on Ec2 is reached, and puppet fails to run, which prevents scaling
and clean runs of Puppet, thus raising the occurrences of swearing.

This work changes the behaviour of the ELB enumeration to first look in
the catalog catalog for the resources that might match, and see if the
relevant data is present before we make the calls to the API.  The
result is that for resources under Puppet management, no extra calls
need to be made to the Ec2 API, thus improving performance and reducing
the risk of reaching the Ec2 API limits.  This also means that there is
an incentive for moving resources under Puppet management, as even those
ELBs that are not managed by Puppet, but are present in the environment
will drag performance down by requiring subsequent calls for the
security groups, subnets, and Ec2 instances related to the particular
ELB.

### DIFF
--- a/lib/puppet/provider/elb_loadbalancer/v2.rb
+++ b/lib/puppet/provider/elb_loadbalancer/v2.rb
@@ -5,60 +5,112 @@ Puppet::Type.type(:elb_loadbalancer).provide(:v2, :parent => PuppetX::Puppetlabs
 
   mk_resource_methods
 
-  def self.instances
+  def self.instances(ref_catalog=nil)
+    Puppet.debug('Fetching ELB instances')
     regions.collect do |region|
-      retries = 0
-      begin
         load_balancers = []
-        region_client = elb_client(region)
-        region_client.describe_load_balancers.each do |response|
-          response.data.load_balancer_descriptions.collect do |lb|
-            load_balancers << new(load_balancer_to_hash(region, lb))
+        elbs do |lb|
+          retries = 0
+          begin
+            load_balancers << new(load_balancer_to_hash(region, lb, ref_catalog))
+          rescue Aws::EC2::Errors::RequestLimitExceeded => e
+            retries += 1
+            if retries <= 8
+              sleep_time = 2 ** retries
+              Puppet.debug("Request limit exceeded, retry in #{sleep_time} seconds")
+              sleep(sleep_time)
+              retry
+            else
+              raise PuppetX::Puppetlabs::FetchingAWSDataError.new(region, self.resource_type.name.to_s, e.message)
+            end
+          rescue Timeout::Error, StandardError => e
+            raise PuppetX::Puppetlabs::FetchingAWSDataError.new(region, self.resource_type.name.to_s, e.message)
           end
         end
-        load_balancers
-      rescue Aws::EC2::Errors::RequestLimitExceeded => e
-        retries += 1
-        if retries <= 8
-          sleep_time = 2 ** retries
-          Puppet.debug("Request limit exceeded, retry in #{sleep_time} seconds")
-          sleep(sleep_time)
-          retry
-        else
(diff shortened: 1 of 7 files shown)


zleswomp:
I'll document the properties I've added here, but feedback here on this style would be welcome.  I know its a bit unconventional, but its sorta fun.

zleswomp:
Brief docs have been added for new properties.  Waiting for test results.

zleswomp:
Now with passing tests.

lib/puppet/provider/elb_loadbalancer/v2.rb
```diff
@@ -5,60 +5,114 @@
...
+              retry
+            else
+              raise PuppetX::Puppetlabs::FetchingAWSDataError.new(region, self.resource_type.name.to_s, e.message)
+            end
+          rescue Timeout::Error, StandardError => e
+            raise PuppetX::Puppetlabs::FetchingAWSDataError.new(region, self.resource_type.name.to_s, e.message)
           end
         end
+
         load_balancers
```
hunner: This looks like it is indented two spaces too far? Maybe it's just github's diffing that is wrong.

lib/puppet/provider/elb_loadbalancer/v2.rb
```diff
@@ -5,60 +5,114 @@
...
       end
-    end.flatten
+    end
   end
 
   read_only(:region, :scheme, :listeners, :tags, :dns_name)
 
   def self.prefetch(resources)
-    instances.each do |prov|
+    ref_catalog = resources.values.first.respond_to?(:catalog) ? resources.values.first.catalog : nil
```
hunner: I know this caused a spec failure, but I think the spec may be wrong... I can't think of any time that prefetch would be called in the real world that WOULDN'T have a populated `resources` hash.

hunner: Yeah, this line explicitly passes an empty hash which would never happen in the real world: https://github.com/puppetlabs/puppetlabs-aws/blob/1.0.0/spec/unit/provider/elb_loadbalancer/v2_spec.rb#L34

lib/puppet/provider/elb_loadbalancer/v2.rb
zleswomp: Good catch.  Fixed.

lib/puppet/provider/elb_loadbalancer/v2.rb
zleswomp: Right, I only put that logic there to satisfy the spec.  Should I make a change here?  I'm not sure how to fake a catalog in the spec when the resources passed are an empty hash.  In practice, this works, but the specs don't generate a catalog that I know how to inspect here.

zleswomp:
@zabacad has raised the point that a page size of 10 is far too small.  I did that to test the logic of making subsequent calls, but is not necessary here.  I've removed the page size entirely to rely on the defaults size.

lib/puppet/provider/elb_loadbalancer/v2.rb
hunner: Meh.

PR #376, opened 2016-10-24
Remove requirement for unique gateway in route table

zvymazal:
It's an incorrect assumption that there's only one network behind a gateway. It must for example be possible to route several networks through VPN gateway.

screen shot 2016-10-24 at 3 11 11 pm

### DIFF
--- a/lib/puppet/provider/s3_bucket/v2.rb
+++ b/lib/puppet/provider/s3_bucket/v2.rb
@@ -44,11 +44,8 @@ Puppet::Type.type(:s3_bucket).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) 
   end
 
   def create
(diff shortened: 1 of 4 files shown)


DoctahPopp871:
👍  we have this commented out internally. @zvymazal you beat me to it !

lib/puppet/provider/s3_bucket/v2.rb
```diff
@@ -44,8 +44,11 @@ def exists?
   end
 
   def create
-    Puppet.debug("Creating S3 Bucket #{name}")
-    s3_client.create_bucket({bucket: name})
+    Puppet.debug("Creating S3 Bucket #{name} in #{target_region}")
```
zvymazal: @zleswomp Sorry, I included some commits that weren't supposed to be added. I'll branch out and create a new pull request.

PR #377, opened 2016-10-25
Modifying logic behind dhcp_options netbios validation

DoctahPopp871:
netbios node type = '2' is not a default value , rather it is the recommended value. Additionally, not all dhcp_options require the netbios_node_type to be specified. After initial provisioning subsequent runs puppet attempts to add the default value 2 for netbios_node_type, only to fail as it is a read-only (creation time only) value. 

This change improves the description of the netbios_node_type property, and enforces the presence of said property when the netbios_name_servers property is in use. netbios_node_type is a requirement when netbios_name_servers is in use. Will fail when both are not specified.

### DIFF
--- a/lib/puppet/type/ec2_vpc_dhcp_options.rb
+++ b/lib/puppet/type/ec2_vpc_dhcp_options.rb
@@ -68,8 +68,7 @@ Puppet::Type.newtype(:ec2_vpc_dhcp_options) do
   end
 
   newproperty(:netbios_node_type) do
-    desc 'The netbios node type, defaults to 2.'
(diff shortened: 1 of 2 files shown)


DoctahPopp871:
adding to unit test to test for new behavior

zleswomp:
+1

DoctahPopp871:
Updated the validation to be a bit more elegant, using self with attribute symbol as key.

hunner:
If someone previously relied on the default of 2 then this would be backwards incompatible for them. We can roll it in as part of the next major bump release when we drop puppet 3.

PR #378, opened 2016-10-25
Remove requirement for unique gateway in route table

zvymazal:
It's an incorrect assumption that there's only one network behind a gateway. It must for example be possible to route several networks through VPN gateway.

24c77282-99fc-11e6-86c2-5b534ab887fe

### DIFF
--- a/lib/puppet/type/ec2_vpc_routetable.rb
+++ b/lib/puppet/type/ec2_vpc_routetable.rb
@@ -47,9 +47,7 @@ Puppet::Type.newtype(:ec2_vpc_routetable) do
   validate do
     routes = self[:routes]
     if routes
-      uniq_gateways = Array(routes).collect { |route| route['gateway'] }.uniq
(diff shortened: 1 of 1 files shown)


hunner:
If you remove this validation and have a route table with multiple gateways, does `puppet resource ec2_vpc_routetable` report the correct gateways, or does it only report a single gateway? If puppet can't understand multiple gateways somewhere else in the module then this failure could be just to safeguard against other more obscure failures or non-idempotent puppet runs.

Thanks!

zleswomp:
@zvymazal Would you mind coming back to address the question here please?

zvymazal:
Hi @zleswomp, I believe it behaves correctly with multiple gateways because that's how we used it, but unfortunately I'm not working on that project anymore and cannot confirm on 100%.

zleswomp:
This seems like a good change.  I don't know how to validate it though.

daveseff:
This may have been taken care of with my PR #505

PR #379, opened 2016-11-04
Add support for security_group changes on ec2_instance

zleswomp:
Without this work, the security_groups parameter on the ec2_instance
type is marked as read-only.  Here we remove the security_groups
parameter from the read-only list and implement a setter method and the
necessary flush logic to modify the security groups attached to an
interface.

This is currently only implemented for instances with a single
interface, as the modification for what security groups are associated
with an instance is a change needed on the interface, and not the
instance.  This is a bit confusing because the interface security groups
are represented on both the main instance object, and the interface
object.  For this reason, a new read-only parameter is also added here
to collect information about the interfaces interfaces to ensure that we
are only modifying the instances we know how, as well as collecting the
data to get the interface ID for modification in the flush method.

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -8,6 +8,11 @@ Puppet::Type.type(:ec2_instance).provide(:v2, :parent => PuppetX::Puppetlabs::Aw
   mk_resource_methods
   remove_method :tags=
 
+  def initialize(value={})
+    super(value)
+    @property_flush = {}
+  end
+
   def self.instances
     regions.collect do |region|
       begin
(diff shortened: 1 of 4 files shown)


zleswomp:
I think the failing tests here are outside the scope of my change.

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -328,11 +350,44 @@ def stop
...
+      # Check if we need to modify the security groups for an interface
+      if @property_flush.keys.include? :security_groups
+        if @property_hash[:interfaces].size != 1
+          Puppet.warning('Handling security group changes on instances with
+                         multiple interfaces is not implemented')
+        else
+          interface_id = @property_hash[:interfaces].keys[0]
+          group_ids = ec2.describe_security_groups({
+            filters: [
+              {name: 'group-name', values: @property_flush[:security_groups]},
```
zabacad: There are other places in this module where various AWS resources can be referenced by ID. Do want to support referencing security groups by ID here?

(I'm thinking no.)

zleswomp: If it turns out that we want it, we can add it later, imo.  Currently, we use security group names, and so here we just extend that a bit to allow changing the attached groups.

lib/puppet/type/ec2_instance.rb
```diff
@@ -235,6 +235,10 @@ def insync?(is)
     end
   end
 
+  newproperty(:interfaces) do
+    desc 'A collected property of the interfaces attached to an instnace'
```
msilvey: 'instance'

PR #383, opened 2016-11-21
Feature/autoscale update

MWilkinson:
Hi, this is a first attempt at adding some of the new functionality to the ec2_autoscalegroup resource.
It allows for the setting of the Termination Policies and Target Groups (feature of elbv2)

### DIFF
--- a/lib/puppet_x/puppetlabs/aws.rb
+++ b/lib/puppet_x/puppetlabs/aws.rb
@@ -162,6 +162,14 @@ This could be because some other process is modifying AWS at the same time."""
         self.class.elb_client(region)
       end
 
+      def self.elbv2_client(region = default_region)
+        ::Aws::ElasticLoadBalancingV2::Client.new(client_config(region))
+      end
+
+      def elbv2_client(region = default_region)
+        self.class.elbv2_client(region)
+      end
+
       def self.autoscaling_client(region = default_region)
         ::Aws::AutoScaling::Client.new(client_config(region))
       end
--- a/README.md
+++ b/README.md
@@ -597,8 +597,14 @@ back- end instances.  Accepts a hash with the following keys:
 ##### `load_balancers`
 *Optional* A list of load balancer names that should be attached to this autoscaling group.
 
+##### `target_groups`
+*Optional* A list of ELBv2 Target Group names that should be attached to this autoscaling group.
+
 ##### `subnets`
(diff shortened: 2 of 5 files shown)


zleswomp:
The readme will need an update with the new properties.

MWilkinson:
Hi @zleswomp, README.md updated as requested

lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -162,6 +162,14 @@ def elb_client(region = default_region)
         self.class.elb_client(region)
       end
 
+      def self.elbv2_client(region = default_region)
```
zleswomp: For my own understanding, what is in the v2 client that we are using that is not present in the ELB client already in the code?

zleswomp:
This looks good to me.  +1

lib/puppet_x/puppetlabs/aws.rb
MWilkinson: elbv2 is the interface to AWS's new application load balancer setup which supports path based routing at the application level. See http://docs.aws.amazon.com/cli/latest/reference/elbv2/index.html

zleswomp: @MWilkinson Looking forward to those new types.  I think I should be able to help test those.

zleswomp: @MWilkinson Have the elbv2 resource taken form yet?  I'm in the middle of adding some code to the elb resource, and oh my, I'm realizing how much needs to be improved.

MWilkinson: @zleswomp it's coming along slowly (seasonal holidays getting it the way) the basic's are there for reporting the targetgroup/loadbalancer
feel free to take a look, I've pushed my current code to a branch here :- https://github.com/ctidigital/puppetlabs-aws/tree/feature/elbv2 (this is branched off the codebase for this pull request)

hunner:
Looks good! Could you also add some tests and squash the commit history? Thanks.

MWilkinson:
Hi @hunner, I've squashed the commit history as requested, but As this only adds the ability to set the target groups on an existing autoscaling group (for elbv2 load balance usage) I'm not sure what tests can be added at this point - as the module currently doesn't support creating elbv2 load balancers (but I'm working on that...)

PR #384, opened 2016-11-28
Attempt fix hash method call on nil object

zleswomp:
Without this change, we call '.keys' on a nil object.  The 'initialize'
method should have set the 'property_flush' to an empty hash, so I'm a
little confused about why this is required.


zleswomp:
@daveseff For your testing please.

hunner:
Yeah, no idea why this would be nil.

daveseff:
Thanks. I will test.

PR #387, opened 2016-12-15
Add support for ELB listener modifications

zleswomp:
Without this change, the listeners on elb_loadbalancer resources are
marked as read-only in the type and are thus not modifiable.  This work
adds support for modifying listeners and setting the 'policy' for HTTPS
listeners to control which cipher suites are available.


zleswomp:
Due to the change in behavior, this warrants an appropriate version bump, as some users may be relying on the notion that puppet will not modify the listeners on their ELBs.  A surprise change here may cause site outages for users that are not prepared for this change.

hunner:
As listeners is "read_only" (write-once at creation, technically) I don't think there would be any existing manifests or workflows that would have different outcomes after this merged, and it would only expand the API of the module, so this should be a feature not a breaking change. Yeah @zleswomp ?

zleswomp:
@hunner Thats right.  If manifests that are deployed with incorrect `listeners`, puppet is unable to change them today.  This code would allow those manifests to proceed, potentially causing a problem.  Just something for release notes.

PR #390, opened 2016-12-15
Update the Gemfile for webmock ruby versions

zleswomp:
Without this change, webmock fails to install correctly due to a version
mismatch on ruby.  Here we address this by pinning the version of
webmock based on the ruby version.


dallinb:
+1

PR #394, opened 2017-01-07
Allows specifying public and private ip address.

kmacoske:
If both public and private ip are specified then the private_ip_address must be within the network interface structure. Otherwise an error is returned from the API request that looks like this:

"Network interfaces and an instance-level private IP address may not be specified on the same request"

### DIFF
--- a/lib/puppet/provider/ec2_instance/v2.rb
+++ b/lib/puppet/provider/ec2_instance/v2.rb
@@ -261,12 +261,7 @@ Found #{matching_groups.length}:
     config
   end
 
-  def config_with_private_ip(config)
-    config['private_ip_address'] = resource['private_ip_address'] if resource['private_ip_address'] && using_vpc?
-    config
-  end
-
-  def config_with_public_interface(config)
+  def config_with_ip(config)
     if resource[:associate_public_ip_address] == :true
       config[:network_interfaces] = [{
         device_index: 0,
@@ -274,9 +269,16 @@ Found #{matching_groups.length}:
         groups: config[:security_group_ids],
         associate_public_ip_address: true,
       }]
+      # If both public and private ip specified, then the private_ip_address must be within the network_interfaces structure
+      if resource['private_ip_address'] && using_vpc?
+      	config[:network_interfaces].first[:private_ip_address] = resource['private_ip_address']
+      end
       config[:subnet_id] = nil
       config[:security_group_ids] = nil
+    elsif resource['private_ip_address'] && using_vpc?
+      config['private_ip_address'] = resource['private_ip_address'] if resource['private_ip_address'] && using_vpc?
     end
+
     config
   end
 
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -261,22 +261,24 @@ def config_with_key_details(config)
...
     if resource[:associate_public_ip_address] == :true
       config[:network_interfaces] = [{
         device_index: 0,
         subnet_id: config[:subnet_id],
         groups: config[:security_group_ids],
         associate_public_ip_address: true,
       }]
+      # If both public and private ip specified, then the private_ip_address must be within the network_interfaces structure
+      if resource['private_ip_address'] && using_vpc?
+      	config[:network_interfaces].first[:private_ip_address] = resource['private_ip_address']
```
hunner: This is oddly indented.

lib/puppet/provider/ec2_instance/v2.rb
```diff
@@ -261,22 +261,24 @@ def config_with_key_details(config)
...
+      end
       config[:subnet_id] = nil
       config[:security_group_ids] = nil
+    elsif resource['private_ip_address'] && using_vpc?
+      config['private_ip_address'] = resource['private_ip_address'] if resource['private_ip_address'] && using_vpc?
```
hunner: duplicated postfix conditional

lib/puppet/provider/ec2_instance/v2.rb
hunner: Also, why just taking the first network interface?

zleswomp:
@kmacoske Would you mind coming back to address the questions here please?

kmacoske:
@zleswomp I certainly will. Thank you for the ping it bumped this pull request back into memory. 

Fixed the indentation and duplicate conditional. 

As for why I'm using only the first interface, it's because currently it seems the module only supports a single network interface. I think that if the module included the ability to specify N number of interfaces, then each should have an optional field for `public_ip_address` and `private_ip_address`. Optional in that if they aren't specified then the API request should be crafted such that AWS will pick an available address. I have no opinion as to how the class paramaters should look if it's only requested to have a single network interface and you only want to specify 'yes' to having a public ip address but still have the ability to specify this for N number of interfaces.  I may not be explaining my thought well, so in a nut shell I just think a slight refactor may be in order to allow the flexability of specifying or not specifying a public and/or private for N number of interfaces. 

I will probably build this functionality out of necessity. If this pull request stays open then I'll add to it when I get the chance to do the work.

zleswomp:
Alright, see.  Does this warrant an update to the README?  Has this been tested in your environment?  I checked out the code and I get the same data structure back when running `puppet resource`.

kmacoske:
I don't think an update to the README regarding the change is necessary. The README currently doesn't state that it does not support specifying `private_ip_address` and `associate_public_ip_address`. Based on that, this is more like a bug fix. 

However an update to the README about the functionality to create multiple interfaces not being implemented may be in order. Although I don't know if it's really that common to want to create multiple interfaces, I just know it's useful for us.

zleswomp:
@kmacoske That sounds good.  I look forward to the multiple interface work, as I think we could use that in a couple places as well.  Thanks for the contribution!

Issue #333, opened 2016-07-29
Implement VCR filter_sensitive_data configuration directive


reubenavery:
Got this tip from the VCR community, there's a facility available to automatically strip secrets from generated vcr_cassettes: https://relishapp.com/vcr/vcr/v/3-0-1/docs/configuration/filter-sensitive-data

zleswomp:
@reubenavery Good find!

ahenroid:
@reubenavery, I believe this was addressed by #399. Can I close this issue?

PR #395, opened 2017-01-10
Begin filtering sensitive data during VCR recording

zleswomp:
Without this change, module testers are left to their own methods to
ensure that they are not leaking any sensitive material when
contributing to this module.  This is not good, as a) we want more tests
and b) we would like not to expose data from contributor environments
when submitting changes.

Here we take a stab addressing this concern by configuring VCR to filter
out the AWS_ACCESS_KEY_ID.  This alone is not enough to be complete, but
it begins the road of making testing simpler on the part of those who
wish to contribute to the module.  Now we can amend the documentation to
include information about exporting the necessary environment variables,
not providers specs don't have to modified last minute to avoid
committing credentials.


zleswomp:
https://github.com/puppetlabs/puppetlabs-aws/issues/333

zleswomp:
I believe these tests should pass after https://github.com/puppetlabs/puppetlabs-aws/pull/397 is merged and a rebase is done here.

hunner:
Rebase away!

zleswomp:
Rebased and updated the tests for the recently merged iam_ types and providers.

lib/puppet/provider/ecs_task_definition/v2.rb
```diff
@@ -1,4 +1,5 @@
 require_relative '../../../puppet_x/puppetlabs/aws.rb'
+require 'pp'
```
zleswomp: Needs removal.

PR #397, opened 2017-01-12
(maint) Comment out iam acceptance tests

hunner:
Our internal AWS testing credentials do not have the ability to manage
iam credentials, but we don't want to delete the tests.

Related: #340


zleswomp:
Woo, would merge if I could.

PR #398, opened 2017-01-14
Add initial support for CloudFront

zabacad:
This adds support for creating, destroying, and enumerating CloudFront web
distributions. Basic parameters and custom origins may be set up. Modification,
several parameters, and S3 origins are left for future work.

### DIFF
--- a/lib/puppet/provider/cloudfront_distribution/v2.rb
+++ b/lib/puppet/provider/cloudfront_distribution/v2.rb
@@ -1,0 +1,245 @@
+require_relative '../../../puppet_x/puppetlabs/aws.rb'
+
+Puppet::Type.type(:cloudfront_distribution).provide(:v2, :parent => PuppetX::Puppetlabs::Aws) do
+  confine feature: :aws
+
+  mk_resource_methods
+
+  read_only(:arn, :id, :status)
+
+  def self.instances
+    dists = []
+    list_opts = {max_items: 100}
+
+    # Loop over paginated API responses.
+    loop do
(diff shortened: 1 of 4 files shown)


zabacad:
I do plan on working on this further, including adding tests.

lib/puppet/provider/cloudfront_distribution/v2.rb
```diff
@@ -0,0 +1,241 @@
...
+      distribution_config: dist_resp.distribution.distribution_config,
+    })
+
+    @property_hash[:enabled] = to_enable
+
+    dist_disabled_resp.etag
+  end
+
+  def flush
+    Puppet.warning('Altering a CloudFront distribution is not supported yet.')
```
hunner: This will be called after create/destroy, in addition to normal property changes. Since `create` updates `@property_hash` you may be able to set like `@property_hash[:just_created] = true` and it will otherwise get ignored by puppet but available for flush to check; I may be wrong on that. For destroy you can compare `if @property_hash[:ensure] == :absent`

hunner:
One comment, but otherwise looks good!

zabacad:
I was seeing the _cannot modify_ warning on creation and destruction. I have confirmed these no longer happen with these changes. Thanks for the help.

Let me know if you want me to squash.

zleswomp:
Squash makes sense here for new functionality.

zabacad:
Squashed.

zleswomp:
This looks good to me.  I'm +1 here for a first pass.

PR #399, opened 2017-01-19
Begin filtering sensitive data during VCR recording

zleswomp:
Without this change, module testers are left to their own methods to
ensure that they are not leaking any sensitive material when
contributing to this module. This is not good, as a) we want more tests
and b) we would like not to expose data from contributor environments
when submitting changes.

Here we take a stab addressing this concern by configuring VCR to filter
out the AWS_ACCESS_KEY_ID. This alone is not enough to be complete, but
it begins the road of making testing simpler on the part of those who
wish to contribute to the module. Now we can amend the documentation to
include information about exporting the necessary environment variables,
not providers specs don't have to modified last minute to avoid
committing credentials.


zleswomp:
Like a pro, I accidentally deleted the remote branch, but this is the same PR as https://github.com/puppetlabs/puppetlabs-aws/pull/395.

zleswomp:
And it passes!

zleswomp:
Hows this look?  Any clarity need adding here?

zleswomp:
I'm holding off sending a PR for https://github.com/puppetlabs/puppetlabs-aws/compare/master...zleswomp:elbcipherpolicies  Once merged here, I'll open a PR based on this work that will include ELB cipher suite management.

zleswomp:
@hunner @DavidS Would either of you mind reviewing here please?  I'd like to get another PR up that is based on this one.

PR #400, opened 2017-01-20
Support for managing EC2 Volumes

daveseff:
I am adding support for managing EC2 Block device volumes. With this you can create and attach volumes to already running instances as well as detach and destroy.

Volumes can be created from scratch or built from a snapshot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,6 +313,7 @@ You can use the aws module to audit AWS resources, launch autoscaling groups in 
 
 * `ec2_instance`: Sets up an EC2 instance.
 * `ec2_securitygroup`: Sets up an EC2 security group.
+* `ec2_volume`: Sets up an EC2 EBS volume.
 * `elb_loadbalancer`: Sets up an ELB load balancer.
 * `cloudwatch_alarm`: Sets up a Cloudwatch Alarm.
 * `ec2_autoscalinggroup`: Sets up an EC2 auto scaling group.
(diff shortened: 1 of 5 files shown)


spec/unit/provider/ec2_volume/v2_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'spec_helper'
+
+provider_class = Puppet::Type.type(:ec2_volume).provider(:v2)
+
+ENV['AWS_ACCESS_KEY_ID'] = 'redacted'
+ENV['AWS_SECRET_ACCESS_KEY'] = 'redacted'
+ENV['AWS_REGION'] = 'sa-east-1'
```
zleswomp: These variables won't need to exist here if https://github.com/puppetlabs/puppetlabs-aws/pull/399 is merged.

zleswomp:
This code looks good to me.  Thank you for putting in the effort!

README.md
```diff
@@ -528,6 +529,35 @@ back- end instances.  Accepts a hash with the following keys:
...
+*Required* The volume type. This can be gp2 for General Purpose SSD, io1 for Provisioned IOPS SSD, st1 for Throughput Optimized HDD, sc1 for Cold HDD, or standard for Magnetic volumes.
+
+##### `encrypted`
+*Optional* Specifies whether the volume should be encrypted. Encrypted Amazon EBS volumes may only be attached to instances that support Amazon EBS encryption. Volumes that are created from encrypted snapshots are automatically encrypted. There is no way to create an encrypted volume from an unencrypted snapshot or vice versa.
+
+##### `kms_key_id`
+*Optional* The full ARN of the AWS Key Management Service (AWS KMS) customer master key (CMK) to use when creating the encrypted volume. This parameter is only required if you want to use a non-default CMK; if this parameter is not specified, the default CMK for EBS is used.
+
+##### `snapshot_id`
+*Optional* The snapshot from which to create the volume.
```
zabacad: This should be *Conditional* to match *size*.

lib/puppet/provider/ec2_volume/v2.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  def destroy
+    Puppet.info("Deleting Volume #{name} in region #{target_region}")
+    # Detach if in use first
+    config = {
+      volume_id: volume_id,
+      force: true
+    }
+    ec2.detach_volume(config) unless @property_hash[:attach] == nil
+    ec2.wait_until(:volume_available, volume_ids: [volume_id])
```
zabacad: Waiting could cause a very long run. I prefer the pattern of attempting the deletion immediately, and if it fails, notify the user to run Puppet again. Check the `cloudfront_distribution` `destroy` method for an example.

zabacad:
None of my comments should stand against a merge since this is a new resource type with no established behaviour.

lib/puppet/provider/ec2_volume/v2.rb
```diff
@@ -0,0 +1,119 @@
...
+    @property_hash[:id] = response.volume_id
+    @property_hash[:ensure] = :present
+  end
+
+  def destroy
+    Puppet.info("Deleting Volume #{name} in region #{target_region}")
+    # Detach if in use first
+    config = {
+      volume_id: volume_id,
+      force: true
```
daveseff: This is in the destroy method of the module. The intention is that this is a volume to be deleted with prejudice. It's actually sparks an idea to have a parameter 'ensure => detached' for those volumes that should be detached with data preserved. If this is a desired feature I can look at that in the future.

zabacad: Okay, since the volume is immediately deleted, there's no reason to care about its data.

zleswomp:
I'm still +1 here.  Merging.

PR #401, opened 2017-01-20
(CLOUD-184, CLOUD-205) Avoid ambiguity when using the AWS_REGION environment var

ahenroid:


ahenroid:
This PR updates and replaces the older, but still relevant https://github.com/puppetlabs/puppetlabs-aws/pull/132. This applies the region policy consistently across all resource types.

@garethr and @DavidS, can I get both of you to sign off on this?

lib/puppet/type/ec2_vpc_dhcp_options.rb
```diff
@@ -17,7 +18,7 @@
     desc 'Tags for the DHCP option set.'
   end
 
-  newproperty(:region) do
+  newproperty(:region, :parent => PuppetX::Property::AwsRegion) do
     desc 'The region in which to assign the DHCP option set.'
     validate do |value|
```
zleswomp: This validate can likely be removed now too.

lib/puppet_x/puppetlabs/property/region.rb
```diff
@@ -0,0 +1,15 @@
+module PuppetX
+  module Property
+    class AwsRegion < Puppet::Property
+      validate do |value|
```
zleswomp: I suspect the failure cases can be cleaned up here a bit.  If we get a value, is it a string, else fail.  If its a string, match a regex to ensure its a valid region.

lib/puppet_x/puppetlabs/property/region.rb
```diff
@@ -0,0 +1,15 @@
...
+        name = resource[:name]
+        fail 'region should not contain spaces' if value =~ /\s/
+        fail 'region should not be blank' if value == ''
+        fail 'region should be a String' unless value.is_a?(String)
+        if !ENV['AWS_REGION'].nil? && ENV['AWS_REGION'] != value
```
zleswomp: The logic is hard for me to read.  Is there another way that this can be written to provide a bit more clarity about the intent?

lib/puppet_x/puppetlabs/property/region.rb
```diff
@@ -0,0 +1,15 @@
...
+          fail "if using AWS_REGION environment variable it must match the specified region value for #{name}"
```
zleswomp: For my own understanding, this is the case because we call *_client methods in the providers, and those use the default region, correct?

ahenroid: @zleswomp Here's the original comment from @garethr: "If using the AWS_REGION environment variable and a manifest where the region property of one or more resources doesn't match you can see unspecified behaviour. See https://github.com/puppetlabs/puppetlabs-aws/pull/117 for discussions."

lib/puppet/type/ec2_vpc_dhcp_options.rb
ahenroid: Ah! Good catch. I missed this one.

ahenroid:
Thanks for the feedback, @zleswomp! I've addressed each of your comments.

lib/puppet_x/puppetlabs/property/region.rb
```diff
@@ -0,0 +1,14 @@
+module PuppetX
+  module Property
+    class AwsRegion < Puppet::Property
+      validate do |value|
+        name = resource[:name]
+        fail 'region should be a String' unless value.is_a?(String)
+        fail 'region should be a valid AWS region' unless value =~ /^(ap-(north|south)east-[12]|ap-south-1|ca-central-1|eu-west-[12]|eu-central-1|sa-east-1|us-(east|west)-[12])$/
```
zabacad: Every time AWS adds a region, this code will need to be changed. I feel a more generic regex would be better.

Ex `[a-z]{2}-[a-z]+-\d`

ahenroid: Good point, @zabacad. I modified the regex

zleswomp:
This looks good to me.  Thanks for bringing this up to date!

PR #404, opened 2017-01-24
Allow security groups mutual peering

zabacad:
Without this change, two security groups cannot peer with each other. The first
security group will be created, but will fail to add rules since the second
security group does not exist yet. Further security groups will not be created
due to the first security group failing and being a dependency. This skips
rules that refer to security groups that do not exist.

### DIFF
--- a/lib/puppet/provider/ec2_securitygroup/v2.rb
+++ b/lib/puppet/provider/ec2_securitygroup/v2.rb
@@ -164,22 +164,25 @@ Puppet::Type.type(:ec2_securitygroup).provide(:v2, :parent => PuppetX::Puppetlab
 
         group_response = ec2.describe_security_groups(filters: filters)
         match_count = group_response.data.security_groups.count
-        msg = "No groups found called #{source_group_name}"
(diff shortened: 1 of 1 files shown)


lib/puppet/provider/ec2_securitygroup/v2.rb
```diff
@@ -191,11 +194,13 @@ def authorize_ingress(new_rules, existing_rules=[])
     to_delete = parser.rules_to_delete(existing_rules)
 
     to_delete.reject(&:nil?).each do |rule|
-      ec2.revoke_security_group_ingress(prepare_ingress_for_api(rule))
+      prepared_rule = prepare_ingress_for_api(rule) and
+        ec2.revoke_security_group_ingress(prepared_rule)
     end
 
     to_create.each do |rule|
```
zleswomp: Should we also reject nil here during the create?

zabacad: Changed both `to_create` and `to_delete` to include `compact`.

zleswomp:
This looks good to me.  Thank you for the patch.

PR #406, opened 2017-01-31
Add role management for ECS tasks

zleswomp:
Without this change, the ECS tasks are unable to assume a role when
under puppet management.  This work adds a new 'role' property that
allows tasks to assume a given IAM role.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1083,8 +1083,6 @@ In the case where a user wishes to remove an option from the container, one of t
 
 It's a small kludge, I know.
 
-
-
 ##### `container_definitions`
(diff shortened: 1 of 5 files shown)


zleswomp:
The test failure here would be addressed by a merge on https://github.com/puppetlabs/puppetlabs-aws/pull/399.

zleswomp:
I've got another that I'll put up after this here, based on this work: https://github.com/zleswomp/puppetlabs-aws/compare/ecstaskrole...zleswomp:ecstaskvolumes

zleswomp:
Unit tests are passing here, but there is some gem issue.

PR #408, opened 2017-02-02
Begin ELB listener policy management

zleswomp:
Without this change, the module does not have the facility to manage the
ELB policies, which include a policy type for managing the cipher suite
accepted on HTTPS listeners.  This work adds the initial functionality
to set the listener polices.


zleswomp:
Travis is ignoring me, but I've asked for rebuild here: https://travis-ci.org/puppetlabs/puppetlabs-aws/builds/197759967

zleswomp:
I've just pushed a fix for an issue we discovered to ensure that we don't loose the ability to manage the certificate_id.  Tested function.  No other change to this PR.

zleswomp:
I've rebased to accommodate the changes in #416.

PR #412, opened 2017-02-06
Adds support for CloudFormation.

dharmabruce:


zleswomp:
Travis no workie.  Closing and reopening to kick the thing.

lib/puppet/type/cloudformation_stack.rb
```diff
@@ -0,0 +1,227 @@
...
+
+  newproperty(:creation_time) do
+    desc 'The time at which the stack was created. (readonly)'
+  end
+
+  newproperty(:description) do
+    desc 'A user-defined description associated with the stack. (readonly)'
+  end
+
+  newproperty(:disable_rollback) do
```
zabacad: `newproperty` has a `:boolean` parameter. Would that be applicable here?

dharmabruce: Yes, good catch. I will fix, test, and push a commit.

lib/puppet/type/cloudformation_stack.rb
```diff
@@ -0,0 +1,223 @@
...
+    newvalue(:UPDATE_ROLLBACK_IN_PROGRESS)
+    newvalue(:UPDATE_ROLLBACK_FAILED)
+    newvalue(:UPDATE_ROLLBACK_COMPLETE_CLEANUP_IN_PROGRESS)
+    newvalue(:UPDATE_ROLLBACK_COMPLETE)
+    newvalue(:REVIEW_IN_PROGRESS)
+  end
+
+  newparam(:template_body) do
+    desc 'Structure containing the template body with a minimum length of 1 byte and a maximum length of 51,200 bytes. For more information, go to Template Anatomy in the AWS CloudFormation User Guide.'
+    validate do |value|
```
zleswomp: Since the template will be JSON (correct?) then maybe it would be worth parsing the string here.

lib/puppet/type/cloudformation_stack.rb
```diff
@@ -0,0 +1,223 @@
...
+  end
+
+  newparam(:template_url) do
+    desc 'Location of file containing the template body. The URL must point to a template (max size: 460,800 bytes) that is located in an Amazon S3 bucket. For more information, go to the Template Anatomy in the AWS CloudFormation User Guide.'
+    validate do |value|
+      fail 'template_url must be a String' unless value.is_a?(String)
+    end
+  end
+
+  validate do
```
zleswomp: Oh neat, I didn't know this would work.

zleswomp:
Overall this looks good.  Do you think we should parse the JSON template?

lib/puppet/type/cloudformation_stack.rb
dharmabruce: CloudFormation added support for YAML last September so templates may be written in either JSON or YAML: https://aws.amazon.com/about-aws/whats-new/2016/09/aws-cloudformation-introduces-yaml-template-support-and-cross-stack-references/

If the type validates a template provided in the template_body parameter it implies it would validate a template referenced by the template_url parameter if that was provided instead. I believe it could be the case that the host running the puppet code wouldn't have permissions to access the template referenced by the template_url. I'm not able to come up with a reasonable use-case but maybe there is one.

Also, the AWS SDK validates the template and puppet reports the syntax/schema errors clearly:
Info: Loading facts
Notice: Compiled catalog for sekret-squirrel-srvr in environment production in 0.04 seconds
Info: Applying configuration version '1486591662'
Error: Template format error: Unrecognized parameter type: Strin
Error: /Stage[main]/Main/Cloudformation_stack[s3-bucket-test]/ensure: change from updated to updated failed: Template format error: Unrecognized parameter type: Strin
Notice: Finished catalog run in 14.29 seconds

lib/puppet/type/cloudformation_stack.rb
dharmabruce: (I cribbed that from the ec2_instance.rb code.)

dharmabruce:
I don't think parsing the JSON or YAML templates adds any value since the SDK performs template syntax and schema validation. But please let me know if I'm missing something.

zleswomp:
The only difference in my mind would be the error you get back, and where that comes from.  Failing a catalog compilation due to bad json seems better than setting all the way to the provider call on the agent to get something back from the AWS API that may not be super intuitive.  Other than that, I agree with your assessment.

dharmabruce:
Ok, great point re: failing catalog compilation vs. agent error. I added JSON parsing for validation to the policy_body parameter. (Note that I am not validating the policy document if it's provided via the policy_url instead of the policy_body parameter. Let me know if I should pursue downloading the s3 and parsing, otherwise I think that's overkill.)

I attempted the following implementation for validating the JSON or YAML for the template_body but it has some problems:
```
  newparam(:template_body) do
    desc 'Structure containing the template body with a minimum length of 1 byte and a maximum length of 51,200 bytes. For more information, go to Template Anatomy in the AWS CloudFormation User Guide.'
    validate do |value|
      fail 'template_body must be a String' unless value.is_a?(String)
      fail 'template_body must be longer than 1 byte' unless value.size > 1
      fail 'template_body must be shorter than 51,200 bytes' unless value.size < 51200
    end

    munge do |value|
      begin
        data = JSON.parse(value)
        JSON.pretty_generate(data)
      rescue
        begin
          data = YAML.load(value)
          YAML.dump(data)
        rescue
          fail('template_body string is not valid JSON nor valid YAML')
        end
      end
    end
  end
```

This seems to function but I've found it introduces ambiguity, may introduce strange errors, and may hide errors. For example, JSON is valid YAML, so if the JSON is only slightly incorrect it can fail the JSON.parse but may still pass the YAML.load/dump and then result in a very strange error from the AWS SDK. In these cases, the AWS SDK provides better validation.

The implementation is likely too simplistic. My web searches and perusal through StackExchange didn't result in a good solution, however. Let me know if you have a solution or a reference for solving this problem.

It's not apples to apples by any means, but I perused through ansible's cloudformation implementation and they do not parse the template before the AWS call:
https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/cloud/amazon/cloudformation.py
(Note that they implemented support for YAML before AWS did. They would convert the YAML to JSON if the template_format parameter specified YAML. Once AWS supported YAML directly, they deprecated the template_format parameter.)

zleswomp:
Cool, that looks better to me.  I'd say just a squash on the commits here, and I'm +1 here.

dharmabruce:
Wonderful, thanks very much! Commits squashed.

zleswomp:
Great, thanks for the contribution!

PR #414, opened 2017-02-08
Adds parameter to rds_instance to specify an option group

daveseff:
This PR add the db_option_group to rds_instance to allow users to specify the option group with their RDS instance creation. 

example:

    db_option_group         => 'custom-sqlserver-se-12-00',

The PR also adds the puppet resource "rds_db_option_group". There is no functionality as of yet to manage the option groups themselves.

### DIFF
--- a/lib/puppet/type/rds_db_option_group.rb
+++ b/lib/puppet/type/rds_db_option_group.rb
@@ -1,0 +1,32 @@
+Puppet::Type.newtype(:rds_db_option_group) do
+  @doc = 'Type representing an RDS DB Option group.'
+
+  newparam(:name, namevar: true) do
+    desc 'The name of the DB Option Group (also known as the db_option_group_name).'
+    validate do |value|
+      fail 'name should be a String' unless value.is_a?(String)
+    end
+  end
+
+  newproperty(:description) do
+    desc 'The description of a DB option group.'
+    validate do |value|
+      fail 'description should be a String' unless value.is_a?(String)
+    end
+  end
+
+  newproperty(:options) do
+    desc 'The additional options to apply to the RDS instance'
+    validate do |value|
+      fail 'description should be a String' unless value.is_a?(String)
+    end
+  end
+
+  newproperty(:region) do
+    desc 'The region in which to create the db_option_group.'
+    validate do |value|
+      fail 'region should be a String' unless value.is_a?(String)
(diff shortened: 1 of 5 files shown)


lib/puppet/type/rds_db_option_group.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  newproperty(:description) do
+    desc 'The description of a DB option group.'
+    validate do |value|
+      fail 'description should be a String' unless value.is_a?(String)
+    end
+  end
+
+  newproperty(:options) do
+    desc 'The options'
```
zleswomp: Knowing how to set this value, or what data type `options` is expecting would help here I think.

lib/puppet/type/rds_db_option_group.rb
```diff
@@ -0,0 +1,29 @@
+Puppet::Type.newtype(:rds_db_option_group) do
```
zleswomp: Since this is a new type, we should document it in the readme.

daveseff: The readme should have the new parameter defined ( in this PR)

lib/puppet/type/rds_db_option_group.rb
daveseff: Agreed. This was done in haste due to it being needed quickly in my environment. I will update this description.

daveseff: Yes, @zleswomp After revisiting my changes, I've realized that I never finished the "options" enumeration in rds_db_option_group. So let's no merge just yet as I do some more . That being said, the feature added to rds_instance looks to be working great.

lib/puppet/type/rds_db_option_group.rb
zleswomp: Ah yes apologies, I just overlooked it.

lib/puppet/type/rds_db_option_group.rb
zleswomp: Good news!  Let me know when its ready for review again.

gregohardy:
Thanks for the contribution Dave. Merging.

sheenaajay:
Thanks for submitting the PR @daveseff . Much appreciated for the new parameter added to the rds_instance provider.

sheenaajay:
merging the changes

PR #415, opened 2017-02-08
Begin KMS management

zleswomp:
Without this change, there is no functionality in this module to manage
the lifecycle and policy of a KMS key in amazon.  Here we add a simple
type and provider with just enough to create, destroy, and manage the
policy.  There are testing updates for the normalization code here,
since the KMS policy exercises data structures that are not exercised
elsewhere, so we extend the use case a bit of the normalize method
calls.


zleswomp:
I expect to come back here with updates to the README, tests, and take a pass over any comments or extra requires I left in while developing.  Also, the updates to the normalize_* code and tests will conflict with what is currently in #408, as the tests are relocated for better.

zleswomp:
I've added documentation, cleaned up old commented code, and added a test for the type.  I believe this is ready for review.

PR #416, opened 2017-02-12
(MAINT) Minor fix for ELB acceptance tests

ahenroid:
Fix repeated test failures due to nil value.

### DIFF
--- a/lib/puppet/provider/elb_loadbalancer/v2.rb
+++ b/lib/puppet/provider/elb_loadbalancer/v2.rb
@@ -298,7 +298,12 @@ Puppet::Type.type(:elb_loadbalancer).provide(:v2, :parent => PuppetX::Puppetlabs
     # which to add, and which need modification.  The modification is a bit of
     # a misnomer, since it really ends up being a delete followed by an add.
 
-    is_listener_ports = @property_hash[:listeners].collect {|x| x['load_balancer_port'].to_i }
+
+    is_listener_ports = if @property_hash[:listeners]
+      @property_hash[:listeners].collect {|x| x['load_balancer_port'].to_i }
+    else
+      []
+    end
     should_listener_ports = resource[:listeners].collect {|x| x['load_balancer_port'].to_i }
 
     # Collect a list of ports for listeners that should be deleted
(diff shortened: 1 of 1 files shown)


ahenroid:
I am attempting to restore the AWS acceptance tests to working order and get the AWS module pipeline green again. The vast majority of the failures were due to AWS resource limits, and I've removed all of the stale AWS resources that have accumulated in sa-east-1. There are also minor code fixes coming along, like this one, to get things running again.

ahenroid:
@hunner If this small change looks reasonable, can you hit "merge"? I'm hoping to get the pipeline going again. Thanks!

lib/puppet/provider/elb_loadbalancer/v2.rb
```diff
@@ -298,7 +298,7 @@ def update_listeners
     # which to add, and which need modification.  The modification is a bit of
     # a misnomer, since it really ends up being a delete followed by an add.
 
-    is_listener_ports = @property_hash[:listeners].collect {|x| x['load_balancer_port'].to_i }
+    is_listener_ports = Array(@property_hash[:listeners]).collect {|x| x['load_balancer_port'].to_i }
```
hunner:
```ruby
[1] pry(main)> a = {a: 1}
=> {:a=>1}
[2] pry(main)> Array(a)
=> [[:a, 1]]
[3] pry(main)> [a] unless a.is_a?(Array)
=> [{:a=>1}]
```

hunner: Though there is also this case, just generally
```ruby
[4] pry(main)> a = nil
=> nil
[5] pry(main)> Array(a)
=> []
[6] pry(main)> [a] unless a.is_a?(Array)
=> [nil]
```

hunner: And even though the `listeners` property is supposed to accept an array according to the type, it accepts an array of hashes according to https://github.com/puppetlabs/puppetlabs-aws#usage but an array of arrays of hashes according to https://github.com/puppetlabs/puppetlabs-aws/blob/master/lib/puppet/type/elb_loadbalancer.rb#L33-L39 . Something is fishy.

hunner: More fun, just because:
```ruby
[1] pry(main)> a = {a: {a: 1}}
=> {:a=>{:a=>1}}
[2] pry(main)> Array(a)
=> [[:a, {:a=>1}]]
[3] pry(main)> a = [{a: 1}]
=> [{:a=>1}]
[4] pry(main)> Array(a)
=> [{:a=>1}]
```

ahenroid:
@hunner I replaced "Array(X)" with "if X", which I think meets your criteria.

ahenroid:
@hunner Does this look right now? I'm happy to do another round if this doesn't do it.

ahenroid:
@garethr Do you want to take a quick look? This is a very minor change, and I'm sure there will be other failures behind this one, but this should at least allow the pipeline to get a bit further. Thanks!

garethr:
@ahenroid looks good to me. I've not had chance to run through but the change feels sensible and the point is to exercise the tests. Merging.

PR #418, opened 2017-02-28
Add support for private Route53 zones

zabacad:
This adds functionality to the *route53_zone* resource type to manage private
Route53 zones. Previously, only public zones could be created and destroyed,
and there were no properties to manage. This also adds properties to discovery
and management, including which VPCs to associate with private zones.

Without these changes, only 100 zones could be managed. This updates the
discovery API call to handle pagination to remove this limitation.

The shared code to discover a VPC name by ID is changed to also allow discovery
of a VPC ID by name. Internal structure is changed, but method parameters and
returns retain previous functionality.

Existing Route53 Zone tests are updated for these changes.


zabacad:
TravisCI is having problems today, but the tests pass locally.

zleswomp:
This work looks good to me.  Thanks for putting in the effort!  Anyone else here want to chime in?

zleswomp:
Yep, still looks good to me.  Merging.

Issue #297, opened 2016-04-15
Documentation might be out of date 


Sher-Chowdhury:
I think the documentation needs updating with respect to how to install gems. See:

https://ask.puppet.com/question/25719/getting-error-could-not-find-a-suitable-provider-error-message-when-puppetlabs-aws-module/?answer=25725#post-id-25725

ahenroid:
Thanks, @Sher-Chowdhury. I'm doing some house cleaning now and submitted https://github.com/puppetlabs/puppetlabs-aws/pull/423 to address this.

zleswomp:
PR #423 has been merged.

ahenroid:
Thanks @zleswomp!

PR #423, opened 2017-03-04
Minor update for README install instructions

ahenroid:


ahenroid:
Remove instructions for old and out-of-date versions of Puppet. This is to fix https://github.com/puppetlabs/puppetlabs-aws/issues/297

PR #424, opened 2017-03-06
Fix bug on ecs_task_definition flush

zleswomp:
Without this change, ecs_task_definitions are not created correctly initially
due to a reference to the @property_hash, which may not be populated.  This
change replaces the use of @property_hash to use when generating a new container
definition.  This should allow new contains to get spun up correctly, while also
allowing existing containers to be modified as desired.


zleswomp:
I'm working through this currently.

zleswomp:
This is not the real issue.  I'll send another PR for the real problem.

PR #425, opened 2017-03-07
Look for truncated results on ecs_task_definition resources

zleswomp:
Without this change, if the task family count is larger than the results AWS
is willing to hand back in a single request, we don't correctly discover all of
the resources.  Here we add the logic we use elsewhere to ensure that all
resources are collected.


zleswomp:
We should really centralize the code we use for this logic.

PR #427, opened 2017-03-09
Properly handle array of domain_name for ec2_vpc_dhcp_options

edestecd:
The domain name should be a space separated list not a comma separated list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,7 +919,7 @@ The type of customer gateway. The only currently supported value --- and the def
 *Optional* The region in which to assign the DHCP option set. For valid values, see [AWS Regions](http://docs.aws.amazon.com/general/latest/gr/rande.html#ec2_region).
 
 #####`domain_name`
-*Optional* The domain name for the DHCP options. This parameter is set at creation only; it is not affected by updates. Accepts any valid domain.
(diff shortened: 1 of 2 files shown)


zleswomp:
Thank you for the PR.  Is there a README update that needs to be made for this change also?

edestecd:
Yea, I should probably add some doc...

edestecd:
@zleswomp Doc added.

zleswomp:
This looks reasonable to me, though, have you been able to test this?  Does it work as desired?

edestecd:
I ran it in our test vpc and created a few dhcp_options. It worked as expected and on successive puppet runs it was happy. There is a spec test to make sure the comparison works as well, so if thats passing I suspect thats a good indication as well.
https://github.com/puppetlabs/puppetlabs-aws/blob/master/spec/unit/type/ec2_vpc_dhcp_options_spec.rb#L58

I'm going to be using it tomorrow to build our production vpc...

edestecd:
Can you think of any additional spec test we might want?

edestecd:
@zleswomp what do you think? Anything else I can add?

zleswomp:
Alright, I read over the docs for this.  This looks good to me.  Thanks for the contribution.

PR #437, opened 2017-03-20
Fix new RDS instance tagging

zabacad:
Without these changes, RDS instance creation breaks on tagging the new
instance. This saves the response from RDS instance creation so the generated
ARN can be used to tag the RDS instance. It removes use of `tags_for_resource`
since that expects the tags property to be named `tags` instead of `rds_tags`.


zabacad:
Rebased for tests.

zabacad:
Closing in favor of #442 .

PR #438, opened 2017-03-20
Drop Puppet Enterprise from metadata

zabacad:
Without this change, tests are failing due to a Puppet Forge change.


wilson208:
Tests are failing due to a change which went in to the metadata-json-lint gem to no longer allow the `pe` requirement in the metadata.json. I am working on MODULES-4556 which deals with this also.

Thanks! 👍

PR #448, opened 2017-04-14
Fix restoring PostgreSQL RDS from snapshot

zabacad:
Without this change, attempting to create a new RDS instance from a snapshot
fails when the engine is PostgreSQL. This adds PostgreSQL to the list of
engines that cannot have the DB name specified when restoring from snapshot.

This also corrects the restore snapshot to be a parameter, not a property.


zleswomp:
Looks good to me.  THanks for the patch.

PR #457, opened 2017-04-27
Improve name resolution method isolation and debug output

zleswomp:
This change moves some of the logic for retrieving object names into their own
method to help readability.  Also added here is improved debug output about why
a call to AWS is being made for each of the objects that fail to resolve from
the catalog.  This should help users identify which objects need to be pulled
into the catalog to improve performance.

### DIFF
--- a/lib/puppet/provider/elb_loadbalancer/v2.rb
+++ b/lib/puppet/provider/elb_loadbalancer/v2.rb
@@ -7,6 +7,7 @@ Puppet::Type.type(:elb_loadbalancer).provide(:v2, :parent => PuppetX::Puppetlabs
 
   def self.instances(ref_catalog=nil)
     Puppet.debug('Fetching ELB instances')
+
     regions.collect do |region|
         load_balancers = []
         elbs do |lb|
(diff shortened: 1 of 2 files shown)


zleswomp:
I think we can improve this slightly.  I'll update the commit.

zleswomp:
I believe this is ready for review.  The diff looks weird for the first commit, but its only moving the logic out to its own methods and updating the debug output to include what the heck we're looking for.  The second commit moves the actual lookup centrally.

There are many more places that we call for subnets, and likely instances too that we can replace with these method calls, but I think we can do that in another PR.  This one is hefty enough as it is.

@hunner If you get some time, I'd appreciate review here.

zleswomp:
Fixed a typo in a method name.  No other changes.

lib/puppet/provider/elb_loadbalancer/v2.rb
```diff
@@ -348,6 +348,8 @@ def availability_zones=(value)
   def security_groups=(value)
     unless value.empty?
       ids = security_group_ids_from_names(value)
+
+      Puppet.debug("Setting IDs #{ids}")
```
zabacad: This debug message could use more context.

lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -358,6 +358,91 @@ def self.vpc_name_from_id(region, vpc_id)
...
+            end
+          end
+        end
+
+        instance_names.collect do |instance_name|
+          @security_groups[region].key(instance_name)
+        end.compact
+      end
+
+      def self.ec2_name_from_id(region, instance_id)
```
zabacad: This should be `ec2_instance_name_from_id`.

PR #461, opened 2017-05-09
Fix subnet name resolution for ELBs

zleswomp:
This work fixes a bug where the subnet_names variable is not populated correctly
due to an incorrect variable reference.  Here we change that variable name to
ensure that the subnet_names are returned correctly.


zleswomp:
Thanks @ahenroid

PR #468, opened 2017-06-26
Downgrade metadata-json-lint for Ruby 1.9

zabacad:
Without this change, tests fail. `metadata-json-lint` is not supported on
Ruby 1.9, but versions of the gem prior to 1.2.0 happen to work for our
purposes. This downgrades `metadata-json-lint` when using Ruby 1.9.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -35,7 +35,8 @@ gem 'retries'
 
 group :development do
   gem 'puppet-lint',                        :require => false
-  gem 'metadata-json-lint',                 :require => false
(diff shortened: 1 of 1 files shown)


zabacad:
I'm not familiar with modulesync, so please let me know if I've goofed this up.

(Really, I'd like to ditch Ruby 1.9 "support").

antgel:
Can someone merge this? It seems to be making all PRs fail in Travis. :(

PR #474, opened 2017-07-31
Provided the folder parth of puppet module

Ganesh533:


gregohardy:
Hi @Ganesh533 It looks like you need to add the modulepath flag back in. This won't work as is. --/etc.....

PR #496, opened 2018-01-22
remove old docs

elimin8r:
README referred to a tests directory which no longer exists. removing the stale documentation to prevent confusion to others.


gregohardy:
Cheers!

Issue #158, opened 2015-05-01
RouteTables with VPC Peering connections fail to parse, blocking creation of other route tables


rkan2002:
The ec2_vpc_routetable provider assumes that the gateway_id of an active route will not be nil.  This results in an exception from the AWS SDK, and causes the entire process to fail.

If a route table includes a VPC Peering connection, the gateway_id will be nil and puppet exits with:

"Error: Could not run: Puppet detected a problem with the information returned from AWS
when looking up ec2_vpc_routetable in us-east-1. The specific error was:

expected params[:internet_gateway_ids][0] to be a string

Rather than report on ec2_vpc_routetable resources in an inconsistent state we have exited.
This could be because some other process is modifying AWS at the same time."

Example output from EC2 client for a route with a peering connection:

<struct destination_cidr_block="10.176.16.0/20", gateway_id=nil, instance_id=nil, instance_owner_id=nil, network_interface_id=nil, vpc_peering_connection_id="pcx-zzzzzz", state="active", origin="CreateRoute">

garethr:
@rkan2002 thanks for reporting this. It's clear from https://github.com/puppetlabs/puppetlabs-aws/issues/185 as well that we don't properly support VPC Peering connections. Based on your explanation we'll try and take a look and fix the issue. Much appreciated. I'll update this issue when we get anywhere.

PR #505, opened 2018-05-02
VPC peering connections (#158)

daveseff:
Adds the ability to configure VPC peering connections in a route table. Fixes issue #158


gregohardy:
Great job. Thanks for the fix Dave.

PR #507, opened 2018-06-11
Add auto value for private IP address

imriss:
When the private_ip_address is set to "auto", it is ignored.


gregohardy:
Thanks for the contribution Reza.

PR #513, opened 2018-08-10
Fix credentials file name in README.md

weyhmueller:
The module tries to retrieve the AWS access credentials from '$::settings::confdir/puppetlabs_aws_credentials.ini' and not from '$::settings::confdir/puppetlabs_aws_configuration.ini', which only stores other module configuration like AWS region or a proxy server url but no credentials.


gregohardy:
Thanks!

PR #529, opened 2019-04-29
(MODULES-9005) Remove regions which are disabled by default

sheenaajay:
This will remove regions which are disabled by default.

Please find the test results

**Test Results**
`with processor Puppet::Reports::Store
      on the ec2_vpc
      on the ec2_vpc_dhcp_options
      on the ec2_vpc_routetable
      on the vpc_subnet
      on the vpc_internet_gateway
      on the ec2_vpc_customer_gateway
      on the ec2_vpc_vpn

Finished in 26 minutes 39 seconds (files took 0.69828 seconds to load)
59 examples, 0 failures`

### DIFF
--- a/lib/puppet_x/puppetlabs/aws.rb
+++ b/lib/puppet_x/puppetlabs/aws.rb
@@ -24,13 +24,24 @@ This could be because some other process is modifying AWS at the same time."""
 
     class Aws < Puppet::Provider
       def self.regions
+        #Remove the new AWS Regions which are disabled by default.
+        disabled_regions = [ "ap-southeast-1", "ap-southeast-2", "ap-northeast-1", "ap-northeast-2", "ap-east1", "ap-south-1"]
+
         if ENV['AWS_REGION'] and not ENV['AWS_REGION'].empty?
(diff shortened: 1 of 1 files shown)


lib/puppet_x/puppetlabs/aws.rb
```diff
@@ -24,13 +24,24 @@ def to_s
 
     class Aws < Puppet::Provider
       def self.regions
+        #Remove the new AWS Regions which are disabled by default.
+        disabledRegions = [ "ap-southeast-1", "ap-southeast-2", "ap-northeast-1", "ap-northeast-2", "ap-east1", "ap-south-1"]
```
tphoney: minor thing, if its a variable it should have look like this_case or is it a constant ?

sheenaajay: @tphoney Thank you for the valuable comment. Updating the file.

sheenaajay: @tphoney Done.Updated the variable name.

